PR #63, opened 2024-06-03
[NayNay] Signing bug for validators

### LINKED ISSUE #77: Display the plaintext signed in the `Sign` Text Input demo flow
The `Sign` Text Input demo flow should show the plaintext that was signed.

rh0delta:
- updated signing with adapters to retry the signing request if a specific error is hit as defined in this issue https://github.com/entropyxyz/sdk/issues/367

closes https://github.com/entropyxyz/sdk/issues/367
close #77


src/flows/sign/index.ts
```diff
@@ -1,7 +1,85 @@
 import inquirer from "inquirer"
+import { u8aToHex } from '@polkadot/util'
 import { initializeEntropy } from "../../common/initializeEntropy"
 import { debug, getSelectedAccount, print } from "../../common/utils"
 
+let msg: string
+let signingAttempts: number = 0
```
mixmix, line 7: :fire: this is a global variable. The way you're using it means that this function will try fixing once, then it will not be reset to 0 after this signing... so all future buggy signings will be skipped.

This function feels like it's way too fancy, and also not really named what it is. I *\think\* all that we need is for the original code to be the same but like this:

```js
const sigOpts = {
  msg: msgParam,
  order: ['deviceKeyProxy', 'noop'],
}

const signature = await entropy.signWithAdaptersInOrder(sigOpt)
  .catch(err => {
    // HACK: sometimes order is reversed D:
    sigOpts.order.reverse()
    return entropy.signWithAdaptersInOrder(sigOpt)
  })

```

mixmix: NOTE this code assumes `signWithAdaptersInOrder` DOES NOT MUTATE the input... that would need checking.
We should not write functions which mutate inputs though :skull:

mixmix: don't see any mutations, should be fine?

mixmix (changes requested):

src/flows/sign/index.ts
rh0delta: > :fire: this is a global variable. The way you're using it means that this function will try fixing once, then it will not be reset to 0 after this signing... so all future buggy signings will be skipped.
> 
> 
> 
> This function feels like it's way too fancy, and also not really named what it is. I *\think\* all that we need is for the original code to be the same but like this:
> 
> 
> 
> ```js
> 
> const sigOpts = {
> 
>   msg: msgParam,
> 
>   order: ['deviceKeyProxy', 'noop'],
> 
> }
> 
> 
> 
> const signature = await entropy.signWithAdaptersInOrder(sigOpt)
> 
>   .catch(err => {
> 
>     // HACK: sometimes order is reversed D:
> 
>     sigOpts.order.reverse()
> 
>     return entropy.signWithAdaptersInOrder(sigOpt)
> 
>   })
> 
> 
> 
> ```

We could do this way of catching the error, but we still need to keep track of the attempts made to signing and handle accordingly. I find the try..catch to be cleaner, thoughts?

rh0delta:
@mixmix ready for re-review

src/flows/sign/index.ts
```diff
@@ -1,7 +1,84 @@
...
+      ({ messageAction } = await inquirer.prompt([messageQuestion]))
+      switch (messageAction) {
+      case 'Text Input': {
+        const { userInput } = await inquirer.prompt([userInputQuestion])
+        msg = Buffer.from(userInput).toString('hex')
+        break
+      }
+      // case 'From a File': {
+      //   break
+      // }
```
frankiebee, line 41: is this needed?

src/flows/sign/index.ts
```diff
@@ -1,7 +1,84 @@
...
+    }
+    const userInputQuestion = {
+      type: "editor",
+      name: "userInput",
+      message: "Enter the message you wish to sign (this will open your default editor):",
+    }
+    // const pathToFileQuestion = {
+    //   type: 'input',
+    //   name: 'pathToFile',
+    //   message: 'Enter the path to the file you wish to sign:',
```
frankiebee, line 28: whats this for?

src/flows/sign/index.ts
mixmix: - `try` / `catch` is the same as `promise.catch` :shrug: (don't mind)
- your global variable is still a liability - if multiple people are calling this function your global variable becomes a race condition

You need to change this to not use it - either the way I suggested, or pass the attempts count as an argument

```js
async function signWithAdaptersInOrder (entropy, signingData, attempts = 0) {
  if (attempt > 1) throw Error('NOPE')
  // ...


  signWithAdaptersInOrder(entropy, signingData, attempt + 1)
}
```

mixmix:
@rh0delta that :fire: is still an issue. I've added more context why it's dangerous
https://github.com/entropyxyz/cli/pull/63/files#r1628548871

rh0delta:
@mixmix this is ready for re-review. I followed your idea of passing in the signing attempts, but rather than having the check for the number of attempts at the top, I've added it to the conditional for retrying the signing request with the shuffled order

mixmix (approved):

PR #73, opened 2024-06-04
milleseconds not seconds for time stamp

frankiebee:
fix the math for time calculations
@johnnymatthews was seeing some time issues so was i so now i buffer the time stamp with time it takes to query and all th math is done in "native unix time stamp numbers" rather then trying to convert to seconds

### DIFF
--- a/src/common/initializeEntropy.ts
+++ b/src/common/initializeEntropy.ts
@@ -4,7 +4,7 @@ import Entropy, { wasmGlobalsReady } from "@entropyxyz/sdk"
 import Keyring from "@entropyxyz/sdk/keys"
 import inquirer from "inquirer"
 import { decrypt, encrypt } from "../flows/password"
-import { TIME_THRESHOLD, debug } from "../common/utils"
+import { debug } from "../common/utils"
 import * as config from "../config"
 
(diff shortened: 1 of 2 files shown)


frankiebee:
this is no longer a fix but a removal 😢 

```
// this is the date from substrate
> new Date(1717519152000).toString()
'Tue Jun 04 2024 06:39:12 GMT-1000 (Hawaii-Aleutian Standard Time)'
//this is the date of my machine 
> new Date(1717537414357).toString()
'Tue Jun 04 2024 11:43:34 GMT-1000 (Hawaii-Aleutian Standard Time)'
```

after checking multiple times. trying to get the number from BN to a working date or something i can do math with the time discrepancy is to large from polkadot js side. and it looks like My Date.now() is correct butb the date object created from the query from polkadot is not. nothing i feel like fixing so just removing it instead. To properly fix this we would need to use the lib BN.js and i think all the conversion is a waste of resources.

frankiebee:
not fixing time stamp because the next release of `entropy-core` will no longer require timestamps see: https://github.com/entropyxyz/entropy-core/pull/866

PR #96, opened 2024-06-12
Vi/xdg conformant config

mixmix:
Fixup PR


src/config/index.ts
```diff
@@ -7,7 +7,7 @@ import path from 'path'
 import { migrations } from './migrations'
 import { debug } from '../common/utils'
 
-const { config: configDir } = envPaths('entropyxyz', { suffix: '' })
+const { config: configDir } = envPaths('entropy-cryptography', { suffix: '' })
```
mixmix, line 10: Vi pointed out that we're already using `entropy-cryptography` for other projects

PR #108, opened 2024-06-13
Mixmix/test migration

mixmix:
write tests over config + migrations

this is a base PR that I will extend the "move config" migration from


src/config/index.ts
```diff
@@ -1,31 +1,57 @@
-import { statSync, readFileSync } from 'node:fs'
+import { /* statSync, */ readFileSync } from 'node:fs'
```
rh0delta, line 1:
```suggestion
import { readFileSync } from 'node:fs'
```

src/config/index.ts
```diff
@@ -1,31 +1,57 @@
...
 import { readFile, writeFile } from 'node:fs/promises'
-import { migrations } from './migrations'
-const configPath = `${process.env.HOME}/.entropy-cli.config`
+// import { join } from 'path'
```
rh0delta, line 3:
```suggestion
```

rh0delta (changes requested):

mixmix:
@rh0delta can you check this, and merge if all good?

rh0delta (approved):

PR #109, opened 2024-06-13
migrate config file to new location

mixmix:
Depends on : https://github.com/entropyxyz/cli/pull/108

This PR moves the config to new location

TODO:
- [x] discuss if we want the location to follow standards, or be uniform across OS


mixmix:
@vitropy we're making another pass at this XDG thing.

There's a question we need to answer which is: **"We have a good pattern for Linux... but what should it be for OSX, Windows?"**

Do you know about where we're putting files for other projects on OSX/Windows? It would be good to get some alignment across teams

The library [env-paths](https://npmjs.com/env-paths) suggests those locations should be:
- **Linux**:
  - `$XDG_CONFIG_HOME/entropy-cryptography/entropy-cli.json`
  - `~/.config/entropy-cryptography/entropy-cli.json` (default)
- **OSX**
  - `~/Library/Preferences/entropy-cryptography/entropy-cli.json`
- **Windows**
  -  `%APPDATA%\entropy-cryptography\Config\entropy-cli.json`
  - `C:\Users\USERNAME\AppData\Roaming\entropy-cryptography\Config\entropy-cli.json` (default)

@johnnymatthews raised a good question which was "Isn't having different location on different OS gonna be bad DX?". I've used these standards in another project, and he's not wrong. But ... standards?
We need to decide which is the higher priority - standards, or cross-platform DX :sob:

johnnymatthews:
In the interest of not bike shedding this into oblivion, let's go with the proposal from @vitropy.

It makes the job of documenting stuff marginally more difficult, but it's not the end of the world. MacOS is already a pain anyway.

mixmix:
Thanks for your input @vitropy @johnnymatthews

Is there somewhere we should document this choice, so that we can help any Entropy projects align with this?
Notably Vi, I think you have scripts which use XDG standards, which would mean your scripts would put files in a different location now that this PR with `env-paths`. Not a huge deal but it would nice to be **together** and aligned with all our tools.

rh0delta (approved):
this looks good but please test it thoroughly to ensure the cli still starts using yarn start or the global install, don't want to run into the esm errors again

mixmix:
we need to write tests for that aye @rh0delta

mixmix:
TODO: merge this when https://github.com/entropyxyz/cli/pull/108 is ready

mixmix:
merge squash is so annoying for branch which branch off branches.
every time it just generates merge conflicts because now the history has diverged. Probably need to discuss this as a team

mixmix:
@rh0delta can you re-review, and if good to go, merge? <3

rh0delta (approved):

PR #136, opened 2024-06-25
[NayNay] Updating testnet endpoint

### LINKED ISSUE #135: update test-net enpoint to start with `wss`
the test-net endpoint on it's next reboot will be a wss endpoint. 
please write a migration for this

PS: @johnnymatthews registered user accounts will be cleared on network reboot. We'll think up a strategy to maybe handle this in code. but just fyi this will possibly cause bugs for users their keys will also not have balance.

rh0delta:
closes #135

### DIFF
--- a/src/config/migrations/01.ts
+++ b/src/config/migrations/01.ts
@@ -5,7 +5,8 @@ export function migrate (data = {}) {
     const migratedData = {
       ...data,
       endpoints: {
-        dev: 'ws://127.0.0.1:9944',
+        // @ts-ignore
+        ...data.endpoints,
         'test-net': 'wss://testnet.entropy.xyz'
       },
       'migration-version': version,
--- a/src/config/migrations/01.ts
+++ b/src/config/migrations/01.ts
@@ -1,0 +1,18 @@
+export const version = '1' // make future versions a number plz
+
+export function migrate (data = {}) {
+  try {
+    const migratedData = {
+      ...data,
+      endpoints: {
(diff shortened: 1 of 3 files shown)


src/config/migrations/01.ts
```diff
@@ -0,0 +1,18 @@
+export const version = '1' // make future versions a number plz
+
+export function migrate (data = {}) {
+  try {
+    const migratedData = {
+      ...data,
+      endpoints: {
+        dev: 'ws://127.0.0.1:9944',
+        'test-net': 'wss://testnet.entropy.xyz'
+      },
```
frankiebee, line 10: Non-blocker, but best practice to extend the original object of endpoint not just overwrite the whole thing

frankiebee (approved):

frankiebee (approved):
Looks good but wait to merge till live

mixmix:
I reckon this goes through the `dev` branch right?
This diff is a bit odd because it's going direct into `main` ?

rh0delta:
> I reckon this goes through the `dev` branch right?
> 
> This diff is a bit odd because it's going direct into `main` ?

I think now that the devops team has left the old endpoint open we no longer need to treat this as a hot fix, and we can now update the endpoint and merge to dev and release as per proper channels. If you want to close this, I can re-open a new PR in the morning (EST).

mixmix:
Already done: https://github.com/entropyxyz/cli/pull/143

PR #140, opened 2024-06-26
[NayNay] Transfer updates + unit testing

### LINKED ISSUE #124: [test] transfer
## transfer

```js
transfer(entropy, { from: address, to: address, amount: number })
=> success (true)
=> fail (error)
```

rh0delta:
closes #124


src/flows/entropyTransfer/transfer.ts
```diff
@@ -0,0 +1,34 @@
+import Entropy from "@entropyxyz/sdk";
+import { TransferOptions } from "./types";
+
+export async function transfer (entropy: Entropy, payload: TransferOptions) {
+  const { from, to, amount } = payload
```
mixmix, line 5: :hot_pepper: potential problems:
- `from` is optional any - does `signAndSend` care ?
- `amount` can be "dog" - does `transferAllowDeath` care ?

**Proposal** : I'd feel a lot safer if these unit-test level functions were **super tight**, i.e. I would have written the type

```ts
export interface TransferOptions { 
  from: string
  to: string
  amount: bigint 
}
```
Basically I'm noticing I don't trust Polkadot given past drama and my response to that is to (perhaps paranoid) nail everything down everywhere. This PR suggests a different approach, which may be fine to, but lets align on an approach we wanna all do (whatever it is)

src/common/initializeEntropy.ts
```diff
@@ -17,10 +17,20 @@ const keyrings = {
   default: undefined // this is the "selected account" keyring
 }
 
-export function getKeyring (address) {
-  if (!address && keyrings.default) return keyrings.default
+export function getKeyring (address?: string, seed?: string) {
```
mixmix, line 20: :hot_pepper: this has technical debt creep vibes. Why would the `seed` be on the `accountData` while the `admin.address` is not?

Should this function ever be able to take nothing in? I feel like the real name of this function should be `getCachedKeyring` in which case being able to pass in no address/seed seems wrong to me (all your inputs are optional).

This is definitely a case of me being a pedant. 

**Proposal:**
1. given this function is ONLY used here, we remove the `export`
2. investigate why `accountData` could have a seed but no address? (different issue/ PR?)

src/config/index.ts
```diff
@@ -7,7 +7,7 @@ import envPaths from 'env-paths'
 
 import allMigrations from './migrations'
 
-const paths = envPaths('entropyxyz', { suffix: '' })
+const paths = envPaths('entropy-cryptography', { suffix: '' })
```
mixmix, line 10: :scream: this change is correct, but suggests there was a reversion of a commit I'd done?
Good catch, and also WTF is going on, am I crazy or is something about merge-squash messing up our history. I feel like this is not the first time we've had something like this go weird

src/flows/entropyTransfer/index.ts
```diff
@@ -29,53 +29,48 @@ export async function entropyTransfer ({ accounts, selectedAccount: selectedAcco
...
+    await transfer(
+      entropy, 
+      {
+        from: entropy.keyring.accounts.registration.pair,
+        to: recipientAddress,
+        amount: formattedAmount
+      }
     )
+    b1.stop()
+    clearInterval(interval)
```
mixmix, line 67: This needs to be called in every outcome, but if there is an error right now it will continue.
I propose extracting a function like

```js
function setupTransferProgress () {
  // ...
  const close = () => {
    clearInterval(interval)
    b1.stop()
  }
  return { close }
}
```

then you can call
```js
cons progress = setupTransferProgress()

// and later *everywhere* it's needed drop the stop e.g. 

catch (err) {
  progress.close()
}
```

mixmix: NOTE: this is not your fault, this is a gotcha I'm spotting and is a refactor.... if you want this could be an issue / PR raised that's seperate

src/config/index.ts
rh0delta: No issue with the commit history, this was my mistake. There was an issue found that didn't allow the cli to start and it was due to my machine not having the entropy-cryptography folder but having the entropyxyz directory. Mild oversight

tests/transfer.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+  t.equal(naynayBalance, 0, 'naynay is broke')
+
+  let charlieBalance = await run(
+    'getBalance (charlieStash)',
+    getBalance(entropy, charlieStashAddress)
+  )
+
+  t.equal(charlieBalance, 1e21, 'charlie got bank')
```
mixmix, line 56: :sunglasses:

tests/transfer.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        from: charlieEntropy.keyring.accounts.registration.pair,
+        to: recipientAddress,
+        amount: BigInt(1000 * 10e10)
+      }
+    )
+    // @ts-ignore
+    t.true(transferStatus.isFinalized, 'Funds transferred successfully')
+  } catch (error) {
+    console.error('error', error);
+  }
```
mixmix, line 71:
```suggestion
  const transferStatus = await run(
    'transfer',
    transfer(entropy, {
      from: charlieEntropy.keyring.accounts.registration.pair,
      to: recipientAddress,
      amount: BigInt(1000 * 10e10)
    })
  )
  // @ts-ignore
  t.true(transferStatus?.isFinalized, 'Funds transferred successfully')
```

mixmix (changes requested):
Great clear test, some nice refactoring, couple suggested tweaks.

The only one I want us all to agree on is:
- https://github.com/entropyxyz/cli/pull/140/files#r1661560792 - how flexy are our atomic functions (team decision, I'll discuss with Frankie when I catch up with her)

The other one about `getKeyring` I'm happy to be flexible on, so long as you consider what I've said and thought about future cost of what you're changing

rh0delta:
@mixmix @frankiebee this is ready for re-review

tests/testing-utils/setup-test.ts
```diff
@@ -2,6 +2,8 @@ import { Test } from 'tape'
 import { Entropy, wasmGlobalsReady } from '@entropyxyz/sdk'
 // @ts-ignore
 import { spinNetworkUp, spinNetworkDown, } from "@entropyxyz/sdk/testing"
+// @ts-ignore
+import Keyring from '@entropyxyz/sdk/keys'
```
frankiebee, line 6: is their an an issue for this on sdk? this dosent look good. `why?`

frankiebee (approved):
Looks good to me! but will wait for @mixmix to re-review.

src/common/progress.ts
```diff
@@ -0,0 +1,32 @@
+import cliProgress from 'cli-progress'
+import colors from 'ansi-colors'
+
+export function setupProgress (label: string): { start: () => void; stop: () => void } {
```
mixmix, line 4: # :kissing_heart: 

It's beautiful

src/flows/entropyTransfer/types.ts
```diff
@@ -0,0 +1,7 @@
+// @ts-ignore
+import { Pair } from '@entropyxyz/sdk/keys'
+export interface TransferOptions {
```
mixmix, line 3: # :heart_eyes:

tests/transfer.test.ts
```diff
@@ -0,0 +1,82 @@
+import test from 'tape'
+import { wasmGlobalsReady } from '@entropyxyz/sdk'
+// WIP: I'm seeing problems importing this?
```
mixmix, line 3: Is it resolved in the SDK that is un-published yet I wonder

mixmix (approved):

mixmix:
Great work @rh0delta 
I'm gonna merge, and I've opened an issue for SDK types importing problems

PR #145, opened 2024-06-27
remove duplicate file

mixmix:
This file is a duplicate of `src/tui.ts`
There might be differences, but this one is not referenced by anything.

I renamed that file, so this is the result of a bad merge.
I am suspicious / don't like our default merge-squash strategy... :eyes:


mixmix:
yeah files are identical apart from how the `config.init` is done slightly different in each :shrug: 
no different

PR #147, opened 2024-07-02
[NayNay] Debug Log

rh0delta:
- install and integrated winston logging library into cli
- allowing us to be able to log any and everything to the respective files saved in the ~/path/to/syslogs/entropy-cryptography/entropy-cli.*.log
- included example use of new logger in the tui flow for balance

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -1,0 +1,126 @@
+import envPaths from 'env-paths'
+import { join } from 'path'
+import * as winston from 'winston'
+
+/**
+ * Winston Base Log Levels for NPM
+ * {
+ *    error: 0,
+ *    warn: 1,
+ *    info: 2,
+ *    http: 3,
+ *    verbose: 4,
+ *    debug: 5,
+ *    silly: 6
+ * }
+ */
+
+export class EntropyLogger {
+  protected context: string
+  protected endpoint: string
+  private winstonLogger: winston.Logger
+
+  constructor (context: string, endpoint: string) {
+    this.context = context
+    this.endpoint = endpoint
+
+    let format = winston.format.combine(
+      // Add timestamp key: { timestamp: 'YYYY-MM-DD HH:mm:ss.SSS' }
+      winston.format.timestamp({
+        format: 'YYYY-MM-DD HH:mm:ss.SSS',
+      }),
+      // If message is instanceof Error, log Error's message property and stack
+      winston.format.errors({ stack: true }),
+      // Allows for string interpolation tokens '%s' in message with splat key values
+      // Ex. { message: 'my message %s', splat: ['test'] } -> { message: 'my message test' }
+      winston.format.splat(),
+      // Uses safe-stable-stringify to finalize full object message as string
+      // (prevents circular references from crashing)
+      winston.format.json(),
+    );
+
+    if (process.env.NODE_ENV === 'test') {
+      format = winston.format.combine(
+        format,
+        winston.format.colorize({ level: true }),
+        winston.format.printf(info => {
+          let message = typeof info.message === 'object' ? JSON.stringify(info.message) : info.message;
+          if (info.stack) {
+            message = `${message}\n${info.stack}`;
+          }
+          return `${info.level}: ${message}`;
+        }),
+      );
+    }
+    const paths = envPaths('entropy-cryptography', { suffix: '' })
+    const DEBUG_PATH = join(paths.log, 'entropy-cli.debug.log')
+    const ERROR_PATH = join(paths.log, 'entropy-cli.error.log')
+    const INFO_PATH = join(paths.log, 'entropy-cli.info.log')
+    const VERBOSE_PATH = join(paths.log, 'entropy-cli.verbose.log')
+
+    this.winstonLogger = winston.createLogger({
+      level: process.env.LOG_LEVEL,
+      format,
+      defaultMeta: { service: 'Entropy CLI' },
+      transports: [
+        new winston.transports.File({ filename: VERBOSE_PATH }),
+        new winston.transports.File({
+          level: 'error',
+          filename: ERROR_PATH
+        }),
+        new winston.transports.File({
+          level: 'info',
+          filename: INFO_PATH
+        }),
+        new winston.transports.File({
+          level: 'debug',
+          filename: DEBUG_PATH,
+        }),
+      ],
+    })
+
+    // If env var is set then stream logs to console as well as a file
+    if (process.env.DEBUG) {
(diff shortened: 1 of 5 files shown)


src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+      // (prevents circular references from crashing)
+      winston.format.json(),
+    );
+
+    if (process.env.NODE_ENV === 'test') {
+      format = winston.format.combine(
+        format,
+        winston.format.colorize({ level: true }),
+        winston.format.printf(info => {
+          let message = typeof info.message === 'object' ? JSON.stringify(info.message) : info.message;
```
mixmix, line 47: Sometimes stringify output being all in one line is brutal, and spreading it over more lines, indented is nice.
Obvs takes up more space, not sure impact on logging style here tho

Code that would do that:
```suggestion
          let message = typeof info.message === 'object' ? JSON.stringify(info.message, null, 2) : info.message;
```

src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+    if (process.env.DEBUG) {
+      this.winstonLogger.add(new winston.transports.Console({
+        format: winston.format.cli()
+      }))
+    }
+  }
+
+  // maps to winston:error
+  public error (description: string, error: Error): void {
+    this.writeLogMsg('error', error?.message, this.context, description, error.stack);
```
mixmix, line 92: :hot_pepper: Sometimes you find an error which is weird an nested and doesn't have `error.message` , it's like `error.error.message`. I think we should just pass the whole object in if we fail to get a message, that way there is more context in the log. I think your logger stringifies... so should be ok?

```suggestion
    this.writeLogMsg('error', error?.message || error, this.context, description, error.stack);
```

src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+          }
+          return `${info.level}: ${message}`;
+        }),
+      );
+    }
+    const paths = envPaths('entropy-cryptography', { suffix: '' })
+    const DEBUG_PATH = join(paths.log, 'entropy-cli.debug.log')
+    const ERROR_PATH = join(paths.log, 'entropy-cli.error.log')
+    const INFO_PATH = join(paths.log, 'entropy-cli.info.log')
+    const VERBOSE_PATH = join(paths.log, 'entropy-cli.verbose.log')
```
mixmix, line 59: Please document in `dev/README.md`

https://www.npmjs.com/package/env-paths#pathslog

src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    this.winstonLogger = winston.createLogger({
+      level: process.env.LOG_LEVEL,
+      format,
+      defaultMeta: { service: 'Entropy CLI' },
+      transports: [
+        new winston.transports.File({ filename: VERBOSE_PATH }),
```
mixmix, line 66: :fire: level defaults to `info`
```suggestion
        new winston.transports.File({ 
          level: 'verbose',
          filename: VERBOSE_PATH
        }),
```

src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+    const DEBUG_PATH = join(paths.log, 'entropy-cli.debug.log')
+    const ERROR_PATH = join(paths.log, 'entropy-cli.error.log')
+    const INFO_PATH = join(paths.log, 'entropy-cli.info.log')
+    const VERBOSE_PATH = join(paths.log, 'entropy-cli.verbose.log')
+
+    this.winstonLogger = winston.createLogger({
+      level: process.env.LOG_LEVEL,
+      format,
+      defaultMeta: { service: 'Entropy CLI' },
+      transports: [
```
mixmix, line 65: my understanding from reading docs is that each of these will record things "at the level set AND below" ?
i.e. level "debug" with have everything levels debug and below (i.e. verbose, info, warn, http, error)

I wonder if we really need 4 levels? I wonder if we can drop `verbose` cos it's included in `debug` ?

src/common/logger.ts
mixmix: actually reading more I'm unclear on docs. But I'd love the level to be explicitly set so no-one has to guess reading this code

src/common/logger.ts
```diff
@@ -0,0 +1,126 @@
...
+      );
+    }
+    const paths = envPaths('entropy-cryptography', { suffix: '' })
+    const DEBUG_PATH = join(paths.log, 'entropy-cli.debug.log')
+    const ERROR_PATH = join(paths.log, 'entropy-cli.error.log')
+    const INFO_PATH = join(paths.log, 'entropy-cli.info.log')
+    const VERBOSE_PATH = join(paths.log, 'entropy-cli.verbose.log')
+
+    this.winstonLogger = winston.createLogger({
+      level: process.env.LOG_LEVEL,
```
mixmix, line 62: I'd like to see all the `process.env` extracted as options that the constructor is called with.
```js
constructor (context: string, endpoint: string, { level: string, isTesting: true }) {
```

That or moving those magic external influences so they are all at the top of the function:
```js
constructor (context: string, endpoint: string) {
  const level = process.env.LOG_LEVEL
  const isTesting = process.env.NODE_ENV === 'test'
```

src/config/index.ts
```diff
@@ -7,7 +7,7 @@ import envPaths from 'env-paths'
 
 import allMigrations from './migrations'
 
-const paths = envPaths('entropyxyz', { suffix: '' })
+const paths = envPaths('entropy-cryptography', { suffix: '' })
```
mixmix, line 10: :cry: why this keep happening?

src/common/logger.ts
mixmix: Alternatively, we make a new CLI method so we can call `entropy paths`
and it logs out all the paths that it's using on this OS:

- config / data
- logging ...

Different issue tho

mixmix:
I smoke tested this and found lots of noise in the logs.
But it's not your fault - I found that we're doing baaad things with our config https://github.com/entropyxyz/cli/issues/160

I made a lil PR to try and chill them out and this help but that issue will also need fixing

mixmix (dismissed):
couple :fire: I'd like to see responded to / addressed

There are some other things which need addressing too:
- installing this everywhere to replace our print/ debug
  - figuring out a nice consistent pattern for that for CLI / TUI modes

This is looking great though, I think it's gonna be a massive help

src/common/logger.ts
rh0delta: Removed the verbose log, makes much more sense to just have info, error and debug. However if we want to reduce the noise even more we can filter the logs so that only debug logs go in the debug file, info to the info file and so on. What do you think?

Issue #158, opened 2024-07-03
types resolution


mixmix:
When we try importing e.g. `"entropyxyz/sdk/keys"` we're seeing a Typescript error

```bash
Cannot find module '@entropyxyz/sdk/keys' or its corresponding type declarations.
```

mixmix:
Some research suggests our SDK `package.json` exports are correct... but our `moduleResolution` is wrong?

> TypeScript only supports exports in --moduleResolution `node16` or `nodenext`
>
> src: https://github.com/microsoft/TypeScript/issues/49971#issuecomment-1192993398

However messing with that seems to bring in other moduleResolution problems ... so likely this is related to the general build-system, tsconfig question of "what's the best ts setup"

PR #159, opened 2024-07-03
remove @ts-ignore for @entropyxyz/sdk

mixmix:
Re: https://github.com/entropyxyz/cli/issues/158

We have a bunch of `@ts-ignore` because importing types from `@entropyxyz/sdk` is not working as we'd like.


mixmix:
This is on hold/ blocked on re-working the SDK + CLI build config

Issue #160, opened 2024-07-03
Config de/serialization


mixmix:
While looking at a logger PR I noticed there was some spaaam coming through, followed it upstream and discovered that our config serialization is a big janky... and likewise deserialization

Here's what a config looks like (safe to steal, this is a testing account):

```json
[...]

PR #161, opened 2024-07-03
chill the logs out!

mixmix:
This PR was made to try and fix the logging noise made by Uint8Arrays, but there is still problem upstream:

https://github.com/entropyxyz/cli/issues/160

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -39,7 +39,7 @@ export class EntropyLogger {
       winston.format.splat(),
       // Uses safe-stable-stringify to finalize full object message as string
       // (prevents circular references from crashing)
-      winston.format.json({ replacer: replacer }),
+      winston.format.json({ replacer }),
     );
 
     if (process.env.NODE_ENV === 'test') {
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -2,6 +2,8 @@ import envPaths from 'env-paths'
 import { join } from 'path'
 import * as winston from 'winston'
 
+import { replacer } from './utils'
+
 /**
  * Winston Base Log Levels for NPM
  * {
@@ -14,6 +16,7 @@ import * as winston from 'winston'
  *    silly: 6
  * }
  */
+
 
 export class EntropyLogger {
   protected context: string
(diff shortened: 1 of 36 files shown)


src/common/logger.ts
```diff
@@ -36,7 +39,7 @@ export class EntropyLogger {
       winston.format.splat(),
       // Uses safe-stable-stringify to finalize full object message as string
       // (prevents circular references from crashing)
-      winston.format.json(),
+      winston.format.json({ replacer: replacer }),
```
rh0delta, line 42:
```suggestion
      winston.format.json({ replacer }),
```

src/common/utils.ts
```diff
@@ -13,16 +13,21 @@ export function stripHexPrefix (str: string): string {
 
 export function debug (...args: any[]) {
   _debug(...args.map(arg => {
-    return typeof arg === 'object'
-      ? JSON.stringify(arg, replacer, 2)
-      : arg
+    return stringify(arg, 2)
   }))
 }
```
rh0delta, line 18: i think we are going to move away from using this utility for logs

mixmix: yeah that can be part of your PR I think
- replace `debug` and `print` functions
- uninstall `debug` module

rh0delta: Sounds good to me!

rh0delta:
@mixmix you think we need a migration to update how we are setting the config through config.set? I believe if we update stringification there we should be able to get away from needing a migration because of how we are subscribing to updates. The data gets refreshed everytime, so it should get fixed once the method is updated with your new replacer method

mixmix:
I'm worried it is desrializing wrong. Frankie said it's likely only working cos we use the seed not the secretKey.
i've got it half written today in a branch (pr later)

rh0delta (approved):

mixmix:
well that PR collapsed down nice!
ready for another review @rh0delta / smoke test

mixmix:
@rh0delta GO TO BED
also thanks for the reviews. AND if you review and approve feel free to merge!  (unless there is some "wait till X is in" note)

:heart: :zzz:

PR #164, opened 2024-07-04
patch sign bug

mixmix:
we had a fix in for the whole "reverse order" bug.
The new `sign.ts` file extracted signing functionality, and the fix got lost a bit (there in TUI, CLI got left out).

SO, this
- :heavy_check_mark: moves the order-fix into the general unit-tested function
- :heavy_check_mark: removes it from TUI/CLI specific code (mmmm cleaner!)

_I've been having trouble getting signatures working, so hard to smoke test_


src/flows/sign/sign.ts
```diff
@@ -11,13 +11,26 @@ function stringToHex (str: string): string {
...
+    .catch(async (err) => {
+      // See https://github.com/entropyxyz/sdk/issues/367 for reasoning behind adding this retry mechanism
+      if (err.message.includes('Invalid Signer') || err.message.includes('Invalid Signer in Signing group')) {
+        // retry signing with a reverse order in the subgroups list
+        // QUESTION: will we need a different strategy to retry when there are >2 things to order
+        content.order = content.order.reverse()
+        return entropy.signWithAdaptersInOrder(content)
+      }
+
+      throw err
```
rh0delta, line 34: I'm not sure if we need this fix here anymore since the retry mechanism has now moved to the sdk, which was the intended destination for the fix from the get-go. @frankiebee can you confirm if we are good to remove this here, I believe we talked about removing this as part of a hotfix.

src/flows/sign/index.ts
```diff
@@ -4,65 +4,6 @@ import { initializeEntropy } from "../../common/initializeEntropy"
 import { debug, getSelectedAccount, print } from "../../common/utils"
 import { signWithAdapters } from './sign'
 
-async function signWithAdaptersInOrder (entropy, msg?: string, signingAttempts = 0) {
```
mixmix, line 7: ok... so this PR needs changing ... to remove all this order stuff completely!

mixmix: DONE

mixmix:
@rh0delta ready for review

rh0delta (approved):

PR #165, opened 2024-07-04
Mixmix/config encoding

mixmix:
addresses: https://github.com/entropyxyz/cli/issues/160


src/config/migrations/02.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function isUI8A (thing) {
+  return thing instanceof Uint8Array
+}
+
+
+function objToUI8A (obj) {
+  const bytes = Object.keys(obj)
+    .map(arrayIndex => obj[arrayIndex])
```
rh0delta, line 39: if this is meant to return an array of values from the object, what about using `Object.values(obj)` instead?

mixmix: Ah good point - because the ordering of the values matters a lot and I'm paranoid!
I meant to add a `.sort` on the keys then use those the to ensure the values are in the right order

mixmix:
Ready for review again @rh0delta

rh0delta (approved):

PR #167, opened 2024-07-11
[NayNay] Entropy Register + Unit Tests

### LINKED ISSUE #149: register
```js
register(entropy, { programModAddress?: string, programData: ProgramInstance[] }): Promise<verifyingKey>
```

- programModAddress can be used to overwrite the admin address that could be used from the entropy keyring

rh0delta:
- added new register pure function
- imported new types from sdk
- added unit test for default register

closes #149

### DIFF
--- a/src/common/masking.ts
+++ b/src/common/masking.ts
@@ -16,9 +16,7 @@ export function maskPayload (payload: any): any {
 
   // maskJSONFields doesn't handle nested objects very well so we'll
   // need to recursively walk to object and mask them one by one
-  for (const [property, value] of Object.entries(clonedPayload)) {
-    console.log(property, typeof value);
-    
+  for (const [property, value] of Object.entries(clonedPayload)) {    
     if (value && typeof value === 'object') {
       if (Object.keys(clonedPayload[property]).filter(key => isNaN(parseInt(key))).length === 0) {
         const reconstructedUintArr: number[] = Object.values(clonedPayload[property])
--- a/src/common/masking.ts
+++ b/src/common/masking.ts
@@ -16,7 +16,7 @@ export function maskPayload (payload: any): any {
 
   // maskJSONFields doesn't handle nested objects very well so we'll
   // need to recursively walk to object and mask them one by one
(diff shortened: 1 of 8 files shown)


src/flows/register/register.ts
```diff
@@ -0,0 +1,28 @@
+import Entropy from "@entropyxyz/sdk";
+import { RegsiterParams } from "./types";
+import { print } from "src/common/utils";
+
+export async function register (entropy: Entropy, params?: RegsiterParams): Promise<string> {
+  let verifyingKey: string
+  try {
+    const registerParams = params?.programModAddress && params?.programData ? { programDeployer: params.programModAddress, programData: params.programData } : undefined
```
mixmix, line 8: This bit is new, untested, and unused functionality.

Propose either comment it out or add a test?

mixmix (changes requested):
Looks good. Just one change requested

General reflections
- Love :heart: :fire: the simplicity and clarity of these pure functions. They're also easy to test = the right path
- The `flow/*/index.js` functions now start to feel like the same boiler plate :card_file_box: :yawning_face:  with many of them
  - take `config` state in, use config to start `entropy`
  - why? surely it should just take `entropy` and that's all done a level higher for each flow

Different PR tho (if that could even work .. it's late here)

rh0delta:
@mixmix this PR is ready for re-review

src/common/masking.ts
```diff
@@ -16,7 +16,7 @@ export function maskPayload (payload: any): any {
 
   // maskJSONFields doesn't handle nested objects very well so we'll
   // need to recursively walk to object and mask them one by one
-  for (const [property, value] of Object.entries(clonedPayload)) {
+  for (const [property, value] of Object.entries(clonedPayload)) {
```
frankiebee, line 19:
```suggestion
  for (const [property, value] of Object.entries(clonedPayload)) {
```

frankiebee (changes requested):

frankiebee (approved):
love tests!

PR #192, opened 2024-07-18
stop asking me if I want to continue

mixmix:
I am sick the guts of being asked if I want to go back to the main menu.
This is a little end-of-day fiddle to see what feels like to not be nagged

I think the only thing I'd change is to print a newline before piling straight into the next question

### DIFF
--- a/src/tui.ts
+++ b/src/tui.ts
@@ -71,12 +71,18 @@ async function main (choices, options, logger: EntropyLogger) {
     const newConfigUpdates = await choices[answers.choice](storedConfig, options, logger)
     // TODO: choices should return type EntropyConfig | String
 
-    if (newConfigUpdates !== RETURN_TO_MAIN) {
+    if (isConfig(newConfigUpdates)) {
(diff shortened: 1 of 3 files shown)


src/tui.ts
```diff
@@ -61,38 +62,24 @@ async function main (choices, options, logger: EntropyLogger) {
...
 
   if (!storedConfig.selectedAccount && answers.choice !== 'Manage Accounts') {
     console.error('There are currently no accounts available, please create or import your new account using the Manage Accounts feature')
   } else {
     logger.debug(answers)
     const newConfigUpdates = await choices[answers.choice](storedConfig, options, logger)
-    if (typeof newConfigUpdates === 'string' && newConfigUpdates === 'exit') {
-      returnToMain = true
-    } else {
+    // TODO: choices should return type EntropyConfig | String
```
mixmix, line 72: I think we should check what's coming out of flows before we just plow it into the `config` and persist it aye?
another PR though

mixmix: added a function `isConfig` to minimally check... can extend later

PR #204, opened 2024-07-31
merge user/dev programs flows

mixmix:
This is preparatory work which merges all program related "flows" into same space

I have also moved the programs wasm to the tests folder ... cos they don't need publishing right?


mixmix:
I have read the CLA Document and I hereby sign the CLA

mixmix:
recheck

tsconfig.json
```diff
@@ -19,7 +19,6 @@
   "exclude": [
     "node_modules",
     "dist",
-    "build",
-    "examples"
+    "tests"
```
mixmix, line 22: This was added to stop TSC freaking out about the use of `import.meta.url` - it was concerned that this wouldn't work in CommonJS.

The alternative would be to start messing with the config `compilerOptions.target` ... maybe `es2020` is needed there? But a quick dabble in that direction strongly put me off (raises other problems)

rh0delta (approved):

PR #207, opened 2024-07-31
[NayNay] File Restructure: Base Class + Balance Restructure

### LINKED ISSUE #208: balance restructure

rh0delta:
## Related Issue(s)

- Resolves #208 

## Proposed Changes

- created new base entropy class to handle the shared nature of the intended use of the class
- restructured balance to be in its own parent directory in the root of src
- created command and util file to contain pure functions and controller functions

## Testing

## Checklist

- [ ] I have added tests.


frankiebee:
notes from draft verbal review:
the base class should just be called `Base`
lets drop controller as a naming scheme

src/balance/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+          balanceInfo[address] = { error: error.message }
+        }
+      }))
+      
+      return balanceInfo
+    } catch (error) {
+      this.logger.error(`There was an error getting balances for [${addresses}]`, error);
+      throw new Error(error.message)
+    }
+  }
```
frankiebee, line 43: im torn on this. I have reservations about not having a pure function here

frankiebee: acctually yeah change this. it dosent need to be a class that extends bas the command class should just pass entropy and this should be a pure function

frankiebee (changes requested):

mixmix:
I'm assuming this doesn't need more review from me @frankiebee ?

Anyone feel free to tag me back in if you want attention

rh0delta:
@mixmix would love your thoughts on this one, specifically the new file structure and the organization/use of the new files

rh0delta:
@frankiebee updated the utils to be pure functions, wanna re-review 😬

src/balance/command.ts
```diff
@@ -0,0 +1,18 @@
+import Entropy from "@entropyxyz/sdk"
+import { BaseCommand } from "../common/base-command"
+import * as BalanceUtils from "./utils"
+import { FLOW_CONTEXT } from "./constants"
```
mixmix, line 4: Questions:
1. why did you split out utils?
    - is it because you don't want logger things in that level?
2. why is `FLOW_CONTEXT` extracted?
    - it seems like that's not gonna be used by anything else, so ... premature optimization
    - can you just put it as a `const` at the top of this file / in the constructor?

src/balance/command.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor (entropy: Entropy, endpoint: string) {
+    super(entropy, endpoint, FLOW_CONTEXT)
+  }
+
+  public async getBalance (address: string) {
+    const balance = await BalanceUtils.getBalance(this.entropy, address)
+
+    this.logger.log(`Current balance of ${address}: ${balance}`, `${BalanceCommand.name}`)
+    
+    return `${balance.toLocaleString('en-US')} BITS`
```
mixmix, line 16: I really don't think this should have `BITS`, I don't even know about the `tolocaleString`.
We want this to be CLI first, so just raw JSON friendly formats like numbers, objects, **not** sentences (number + unit).

I could be wrong!

mixmix (changes requested):
Good start to to restructure. I think we need that `flows/balance` folder gone to see how this is gonna land. I think this might end up then touching the root `cli.ts` and `tui.ts`.

And then the whole session concept ....phew

src/balance/command.ts
rh0delta: > Questions:
> 
> 1. why did you split out utils?
>    
>    * is it because you don't want logger things in that level?
> 2. why is `FLOW_CONTEXT` extracted?
>    
>    * it seems like that's not gonna be used by anything else, so ... premature optimization
>    * can you just put it as a `const` at the top of this file / in the constructor?

1. splitting out utils was a decision made to have one unified file that contains the pure functions for the individual command, and remove any idea about having to do anything but make the request to retrieve balance (in this case) and return that data. the removal of logging here was a decision made to allow the command class take care of the "business logic" as my corporate colleagues would call it, which includes error handling/logging and logging of generic information.
2. yeah it was kinda premature, i was under the impression there would be more shareable variables, and originally there was a use since we had the need for flow context in both the utils and command classes, but that has since changed so i will update and remove this file

src/balance/command.ts
rh0delta: the utils method is the pure function that returns a number, my vision was that the controller/command file will handle how we want the data to be presented

rh0delta:
@mixmix got some time for a re-review on this one?

frankiebee (approved):

PR #213, opened 2024-08-01
add deploy to CLI, including start of CLI refactor

mixmix:
This adds deploy to the CLI

This PR also start to add a new pattern for how to CLI which is "one file per major command"
It shows how we're gonna do subcommands too

- [N/A] I have added tests.


mixmix:
@rh0delta I reckon the pattern here for cli could really help split out the tui/cli clutter.

rh0delta:
> @rh0delta I reckon the pattern here for cli could really help split out the tui/cli clutter.

Yeah i agree, i do like the degree of separation here, but will this be the same as how we want to handle the tui as well?

src/cli/commands/entropy-program.ts
```diff
@@ -0,0 +1,70 @@
...
+      const entropy = await getEntropy(opts)
+
+      const pointer = await deployProgram(entropy, {
+        bytecodePath,
+        configurationSchemaPath,
+        auxillaryDataSchemaPath
+      })
+      cliWrite(pointer)
+
+      process.exit(0)
```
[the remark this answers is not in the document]
mixmix: Lord. Good catch!

I think your point about extracting entropy instantiation is a good one. I can think of a nice way to do this I think. I was leaving that level of work out of this PR, but glad you've been working on that.

SO... checking proposed plan with this PR:
- you tell me where that "entropy initialization" refactor is
  - I help you get that branch merged into `dev` (if not already)
- I work on merging new `dev` into this PR?

mixmix:
I'm not sure how the TUI will end up....
I imagined functions that would call commands?

src/cli/commands/entropy-program.ts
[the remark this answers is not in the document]
mixmix: cc @rh0delta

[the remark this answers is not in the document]
mixmix: oh you did show me that. I guess I want to help get those other PRs merged into `dev`

mixmix:
TODO:
- try merge in / rebase #219 
- if that doesn't work, just pull in the BaseCommand class

src/cli.ts
```diff
@@ -120,14 +75,22 @@ program
     launchTui(entropy, options)
   })
 
+/* Install commands */
+// entropyAccount(program)
+// entropyBalance(program)
+// entropySession(program)
+// entropySign(program)
+entropyProgramCommand(entropy, program) // WARNING this has been written to need access to entropy... but that's not defined yet?
```
mixmix, line 83: @rh0delta TODO

src/cli.ts
```diff
@@ -104,6 +57,8 @@ program
       .env('DEV_MODE')
       .hideHelp()
   )
+  // NOTE: I think this hook is gonna need to be on every command? TO BE TESTED
+  // If it is, then extract the hook fn, and loadEntropy out into src/common/utils-cli.ts
```
mixmix, line 61: @rh0delta TODO

src/program/command.ts
```diff
@@ -0,0 +1,56 @@
+import { Command } from 'commander'
+import { Entropy } from '@entropyxyz/sdk'
+import { aliasOrAddressOption, cliWrite, endpointOption } from '../common/utils-cli'
+
+import { deployProgram } from '../flows/programs/deploy'
```
mixmix, line 5: NOTE: this PR was all about setting new pattern for CLI, so moving this into `main.ts` needs to be done later

src/cli.ts
rh0delta: the hook doesnt need to be on every command, just needs to sit at the root command `entropy` and it gets hit for every action

src/cli.ts
rh0delta: entropy gets defined once the cli is initialized, from the root entropy command hook

rh0delta (approved):

src/cli.ts
```diff
@@ -33,10 +34,14 @@ program
...
-    const { account, endpoint } = options
-    const entropy = await loadEntropy(account, endpoint)
-    launchTui(entropy, options)
+  .addCommand(entropyProgramCommand())
+  .action(async (opts: EntropyTuiOptions) => {
+    const { account, endpoint } = opts
+    const entropy = account
+      ? await loadEntropy(account, endpoint)
+      : undefined
+    // NOTE: on initial startup you have no account
```
mixmix, line 43: YO so @rh0delta I had to change this bit cos on initial startup you have no account.
Tested it and it seems to handle it fine ... but do we need to review all the interactions to see if they handle 
"entropy undefined" fine. A couple I tested just said "he you need an account to proceed"

mixmix: I'm gonna merge and raise an issue

mixmix: after tests pass...

mixmix:
Hey @rh0delta can I have a re-review? 
In particular I changed the config handling because ... tests were failing and I didn't like how `get` was having side effects (note it was actually `getAndInitialize` ... which made it quite hard to test and work with.

Through some closer looking I discovered that `config.init` was not actually being called anywhere! Now it is

src/config/index.ts
```diff
@@ -58,33 +59,30 @@ export async function init (configPath = CONFIG_PATH, oldConfigPath = OLD_CONFIG
...
-  try {
-    const configBuffer = readFileSync(configPath, 'utf8')
-    return deserialize(configBuffer)
-  } catch (err) {
-    return makeGetErrorHandler(configPath)(err)
-  }
+export function getSync (configPath = CONFIG_PATH) {
+  const configStr = readFileSync(configPath, 'utf8')
+  return deserialize(configStr)
 }
```
mixmix, line 67: Here is where I change `get` and `getSync` back to good ol' "one purpose" functions

src/cli.ts
```diff
@@ -1,42 +1,52 @@
...
+    launchTui(entropy, opts)
+  })
+  .hook('preAction', async () => {
+    // set up config file, run migrations
+    return config.init()
```
mixmix, line 49: Here is where I ensure `config.init` is always called (I hope?).

This makes sure:
- there is a config in place for when `config.get` / config.getSync` are called
- migrations are run :facepalm:

src/common/utils-cli.ts
```diff
@@ -1,14 +1,23 @@
...
 
+function getConfigOrNull () {
+  try {
+    return config.getSync()
+  } catch (err) {
+    if (config.isDangerousReadError(err)) throw err
+    return null
+  }
+}
+
```
mixmix, line 20: I found this file was the only place `config.getSync` was used. I decided to write a special function to catch any errors and then carefully handled all the cases where it comes back `null`

src/cli.ts
```diff
@@ -1,42 +1,52 @@
...
-  )
+  // NOTE: currently unused
+  // .addOption(
+  //   new Option(
+  //     '-d, --dev',
+  //     'Runs entropy in a developer mode uses the dev endpoint as the main endpoint and allows for faucet option to be available in the main menu'
+  //   )
+  //     .env('DEV_MODE')
+  //     .hideHelp()
+  // )
```
rh0delta, line 33: this option will be used for the faucet, may not want to remove it right now.

src/cli.ts
```diff
@@ -1,42 +1,52 @@
...
+    // NOTE: on initial startup you have no account
+    launchTui(entropy, opts)
+  })
+  .hook('preAction', async () => {
+    // set up config file, run migrations
+    return config.init()
   })
 
-program.parseAsync().then(() => {})
+program.parseAsync()
```
rh0delta, line 52: the chained method was a recommendation by commander, why is it being removed?

rh0delta (changes requested):
couple of questions, and some changes may be needed, otherwise looks good so far

rh0delta:
> Hey @rh0delta can I have a re-review? In particular I changed the config handling because ... tests were failing and I didn't like how `get` was having side effects (note it was actually `getAndInitialize` ... which made it quite hard to test and work with.
> 
> Through some closer looking I discovered that `config.init` was not actually being called anywhere! Now it is

How was it that config.init was never called? really curious about this. I thought we had it initialized at the same point as initializing wasm.

src/cli.ts
mixmix: because it doesn't do anything.
:heavy_check_mark: if it works don't fix it!
:heavy_check_mark: prune dead code?

# :scream: 

not sure which wins. This has been annoying me for a while tho

rh0delta: If there is no significant impact to removing it, I'm cool with it being gone. Have you tested the programmatic cli without the chained method?

src/cli.ts
mixmix: reverted!

src/cli.ts
mixmix: yes!

mixmix:
Yeah I don't know what happened! I assume it just got lost in the refactoring.

src/program/command.ts
```diff
@@ -0,0 +1,62 @@
...
+export function entropyProgramCommand () {
+  return new Command('program')
+    .description('Commands for working with programs deployed to the Entropy Network')
+    .addCommand(entropyProgramDeploy())
+    // TODO:
+    // .addCommand(entropyProgramGet())
+    // .addCommand(entropyProgramListDeployed())
+    // .addCommand(entropyProgramAdd())
+    // .addCommand(entropyProgramRemove())
+    // .addCommand(entropyProgramList())
```
mixmix, line 15: I wanna do these in a different PR. Maybe even after we do a new release.
This is "feature completeness", not "refactor"

src/program/main.ts
```diff
@@ -0,0 +1,67 @@
...
+  async get (programPointer: string): Promise<any> {
+    this.logger.debug(`program pointer: ${programPointer}`, `${FLOW_CONTEXT}::PROGRAM_PRESENCE_CHECK`);
+    return this.entropy.programs.dev.getProgramInfo(programPointer)
+  }
+
+  async listDeployed () {
+    const address = this.entropy.keyring.accounts.registration.address
+    // QUESTION: will we always be wanting this address?
+    return this.entropy.programs.dev.get(address)
+  }
```
mixmix, line 65: NOTE: I think these are about to change with the next SDK release (just reviewing these)

rh0delta (approved):

PR #218, opened 2024-08-07
[NayNay] File Restructure: Signing Restructure

### LINKED ISSUE #212: signing restructure

rh0delta:
## Related Issue(s)

- Resolves #212 

## Proposed Changes

- created new file structure for signing flow
- updated tui/cli with new changes

## Testing

- [ ] Manual testing performed

## Checklist

- [ ] I have added tests.


src/signing/command.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.logger.log(`Signature: ${signatureHexString}`)
+
+      return { signature: signatureHexString, verifyingKey: this.entropy.signingManager.verifyingKey }
+    } catch (error) {
+      this.logger.error('Error signing message', error)
+      throw error
+    }
+  }
+
+  public async runInteraction (inquirer) {
```
mixmix, line 55: I would like to not see Tui mixed in with CLI commands. I thought commands meant "commands that are called by the CLI"

I am keen on CLI commands being:
- the primary thing
- individual functions where the CLI names map pretty directly into our command functions
    - e.g. CLI  `entropy-sign arg1 arg2` => Function `entropySign.default(arg1, arg2)`
    - hmmm interesting , this doesn't really work that well as this one doesn't have a sub-command like other commands. hence suggesting calling it `default` (or something similar?) 

ANYWAY - I'd strongly like to see TUI stuff remove from this file so that it is forever and always a _consumer_ of the CLI commands. Then there can be no accidental problems, and `inquirer` doesn't need to be an argument?

src/signing/command.ts
```diff
@@ -0,0 +1,75 @@
+import Entropy from "@entropyxyz/sdk"
```
mixmix, line 1: Overall, LOVE the pattern this file sets up

src/signing/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+      return
+    }
+  }
+  return {msg, msgPath};
+}
+
+export function getMsgFromInputOrFile (msg?: string, msgPath?: string) {
+  let result: string = msg
+  if (!msg && msgPath) {
+    result = readFileSync(msgPath, 'utf-8')
```
mixmix, line 75: :fire: again, recommend revert.

Problem: What if the file is not utf-8 encoded and is e.g. an `.mp3` file. I *think* this needs some design

mixmix (changes requested):
I love the direction and how you've split out `commands` + `types` + `utils` really clearly.

I've listed quite a few smallish things. Sorry it was 2 weeks and now a bunch of change requests, appreciate that might land a bit annoying.
Happy to have a call if you wanna connect + discuss anything (nicer way to resolve opinions sometimes, and can be faster)

src/signing/command.ts
```diff
@@ -0,0 +1,45 @@
...
+  const programCommand = rootCommand.command('sign')
+    .description('Commands for working with signing with the Entropy Network')
+
+  entropySign(entropy, programCommand)
+}
+
+function entropySign (entropy: Entropy, programCommand: Command) {
+  programCommand.command('sign')
+    .description('Sign a message using the Entropy network. Output is a signature (string)')
+    .argument('msg', 'Message or Path to Message you would like to sign')
```
mixmix, line 16: :fire: this is not feasible. Or rather I cannot see a way you can **know** whether a string is for a path or the actual message content. Examples
- `.env`
- `id_ed25519.pub`
- `https://matou.nz`

Please just make it one or the other `msgPath` or `msg`, not both.
Probably open a design issue too?

src/signing/command.ts
```diff
@@ -0,0 +1,45 @@
...
+    .addOption(endpointOption())
+    .addOption(currentAccountAddressOption())
+    .addOption(
+      new Option(
+        '-r, --raw',
+        'Signs the provided message using the Raw Signing method. Output is a signature (string)'
+      )
+    )
+    .action(async (msg, opts) => {
+      if (opts.accountAddress) {
```
mixmix, line 27: :fire: with `commander` I think it's `opts.account`  (it pays attention to `--account` not `<accountAddress>` ? which is weird to me but I think I tested it and was like "oh dang what?"

src/signing/constants.ts
```diff
@@ -0,0 +1,37 @@
+export const FLOW_CONTEXT = 'ENTROPY_SIGNING'
```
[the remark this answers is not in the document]
mixmix: Still relevant.
Want the folder to reflect the command name (sign, not signing)

src/signing/interaction.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function entropySign (entropy: Entropy, endpoint: string) {
+  const Sign = new EntropySign(entropy, endpoint)
+  const { interactionChoice } = await inquirer.prompt(interactionChoiceQuestions)
+  switch (interactionChoice) {
+  case 'Raw Sign': {
+    const { msg, msgPath } = await getMsgFromUser(inquirer)
+    const { hashingAlgorithm, auxiliaryDataFile } = await inquirer.prompt(rawSignParamsQuestions)
+    let hash = hashingAlgorithm
+    const auxiliaryData = JSON.parse(readFileSync(auxiliaryDataFile).toString())
```
mixmix, line 16: I think this is the same?

```suggestion
    const auxiliaryData = JSON.parse(readFileSync(auxiliaryDataFile, 'utf-8'))
```

src/signing/main.ts
```diff
@@ -0,0 +1,53 @@
+import Entropy from "@entropyxyz/sdk"
+import { u8aToHex } from '@polkadot/util'
+import { EntropyBase } from "../common/entropy-base";
+import { getMsgFromInputOrFile, rawSign, signWithAdapters } from "./utils";
+import { SignResult } from "./types";
+import { FLOW_CONTEXT } from "./constants";
+
+export class EntropySign extends EntropyBase {
```
mixmix, line 8: :fire: please remove the knowledge of `msgPath` from this class. Do that processing upstream, all this class should care about is signing messages that have been passed in.

We want to contain knowledge about all that to one area only, not spread it across layers of our stack.

src/signing/main.ts
```diff
@@ -0,0 +1,53 @@
...
+      )
+      const signatureHexString = u8aToHex(signature)
+      this.logger.log(`Signature: ${signatureHexString}`)
+
+      return { signature: signatureHexString, verifyingKey: this.entropy.signingManager.verifyingKey }
+    } catch (error) {
+      this.logger.error('Error signing message', error)
+      throw error
+    }
+  }
```
mixmix, line 35: I this a new feature? => out of scope

If it's in scope, then I want to see it:
- [ ] plugged into CLI (currently it's commented out)
- [ ] tested

```suggestion
```

src/signing/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  return entropy.signWithAdaptersInOrder({
+    msg: {
+      msg: stringToHex(input.msg)
+    },
+    // type
+    order: ['deviceKeyProxy', 'noop'],
+    signatureVerifyingKey: input.verifyingKey
+    // auxillaryData
+  })
+}
```
mixmix, line 107: move these into `main.ts` please
I'd like to see our tests calling on `main.ts` as that is kinda designed to be the backbone of all our workflows.
I want to know CLI + TUI work with what they are calling (`main.ts` functions)

src/transfer/main.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const FLOW_CONTEXT = 'ENTROPY_TRANSFER'
+
+export class EntropyTransfer extends EntropyBase {
+  constructor (entropy: Entropy, endpoint: string) {
+    super(entropy, endpoint, FLOW_CONTEXT)
+  }
+
+  public async sendTransfer (toAddress: string, amount: string) {
+    const { start: startProgress, stop: stopProgress } = setupProgress('Transferring Funds')
```
mixmix, line 14: :fire: we can't have progress in the programmatic CLI - that could be used in programs which are reading the output of the command. Move this to the TUI

src/transfer/command.ts
```diff
@@ -1,51 +1,21 @@
...
+export async function entropyTransferCommand (entropy: Entropy, rootCommand: Command) {
+  rootCommand.command('transfer')
+    .description('Transfer funds between two Entropy accounts.') // TODO: name the output
+    .argument('source', 'Account address funds will be drawn from')
+    .argument('destination', 'Account address funds will be sent to')
+    .argument('amount', 'Amount of funds to be moved')
+    .addOption(passwordOption('Password for the source account (if required)'))
+    .addOption(endpointOption())
+    .addOption(currentAccountAddressOption())
+    .action(async (_source, destination, amount, opts) => {
```
mixmix, line 15: :fire::fire::fire::fire::fire: why do you drop the `source` here ... if I put in that I wanted to move funds from `A => B` and instead it moved from `selectedAccount => B` I would be wild mad.

so this might need more design... as the options I can see are:
- A. remove `source`
- B. require `source` but then ... have that override the `selectedAccount`   (error if `--account` is provided too?)

src/transfer/utils.ts
```diff
@@ -1,5 +1,32 @@
...
+  name: TRANSFER_CONTENT.recipientAddress.name,
+  message: TRANSFER_CONTENT.recipientAddress.message,
+}
+
+export const transferInputQuestions = [
+  amountQuestion,
+  recipientAddressQuestion
+]
 
 export async function transfer (entropy: Entropy, payload: TransferOptions): Promise<any> {
```
mixmix, line 31: Please move this into `main.ts`

mixmix (changes requested):
## Review 2.

Looking good. There are a few things that are style things (like I want the core functionality all in the `main.ts` file, not hidden elsewhere, and variable renames), and there are also so gotchas which will cause problems :fire: 

The scope creep with new features is starting to burn a bit much.
- I don't want to say again that `msg/msgPath` doesn't work needs more design
- `rawSign` is also new (I think)

**please revert these**. We need this mega branch merged and I don't want to stop to discuss those things :pray: :heart: 
I am real happy to revisit those as simple little side quests after this is merged

---

NOTE: this PR currently doesn't work with fresh installs!  (no accounts = no default selected nor selectable account = throws)

src/signing/command.ts
mixmix: discussed on call, resolved

src/transfer/utils.ts
rh0delta: why?

src/transfer/command.ts
rh0delta: the source in this case must be a funded entropy account, since we require the signer pair, not just an address. The source in this case must be provided by the selected account from the config which can be overrided by the current account address option, so i dont see the need for a specific source argument here

mixmix:
TODO:
- fix `loadEntropy`
- final review

src/common/utils-cli.ts
```diff
@@ -0,0 +1,105 @@
...
+  entropy = await initializeEntropy({ keyMaterial: selectedAccount.data, endpoint, password })
+
+  if (!entropy?.keyring?.accounts?.registration?.pair) {
+    throw new Error("Signer keypair is undefined or not properly initialized.")
+  }
+
+  return entropy
+}
+
+export async function reloadEntropy (entropy: Entropy, newAddress: string, oldAddress: string, endpoint: string): Promise<Entropy> {
```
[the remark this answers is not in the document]
mixmix: resolved: https://github.com/entropyxyz/cli/pull/220

src/common/utils-cli.ts
```diff
@@ -0,0 +1,105 @@
...
+  try {
+    entropy = await loadEntropy(entropy, newAddress, endpoint)
+  } catch (error) {
+    if (error.message.includes('AddressError')) {
+      entropy = await loadEntropy(entropy, oldAddress, endpoint)
+      return entropy
+    }
```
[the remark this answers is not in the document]
mixmix: resolved: https://github.com/entropyxyz/cli/pull/220

src/balance/command.ts
```diff
@@ -1,18 +1,19 @@
...
+export async function entropyBalanceCommand (entropy: Entropy, rootCommand: Command) {
+  rootCommand.command('balance')
+    .description('Command to retrieive the balance of an account on the Entropy Network')
+    .argument('address', 'Account address whose balance you want to query')
+    .addOption(passwordOption())
+    .addOption(endpointOption())
+    .action(async (address, opts) => {
+      const BalanceService = new EntropyBalance(entropy, opts.endpoint)
+      const balance = await BalanceService.getBalance(address)
+      cliWrite(`${balance.toLocaleString('en-US')} BITS`)
```
mixmix, line 15: I don't know if we want ` BITS` in there... but lets merge and see

src/sign/command.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function entropySignCommand (entropy: Entropy, rootCommand: Command) {
+  const signCommand = rootCommand.command('sign')
+    .description('Commands for working with signing with the Entropy Network')
+
+  entropySign(entropy, signCommand)
+}
+
+function entropySign (entropy: Entropy, signCommand: Command) {
+  signCommand.command('sign')
```
mixmix, line 14: :fire: this is a bug I named in the last review. I'm gonna fix it myself

TL;DR you have defined a cli command `entropy sign sign ...`

src/sign/command.ts
```diff
@@ -0,0 +1,40 @@
...
+      // TO-DO: Add ability for raw signing here, maybe? new raw option can be used for the conditional
+      /**
+       * if (opts.raw) {
+       *   implement raw sign here
+       * }
+       */
+      const { verifyingKey, signature } = await SigningService.signMessageWithAdapters({ msg })
+      cliWrite(`Message Signed: ${msg}`)
+      cliWrite(`Verifying Key: ${verifyingKey}`)
+      cliWrite(`Signature: ${signature}`)
```
mixmix, line 37: Discussed on call - we're gonna make this an object output

src/transfer/main.ts
```diff
@@ -0,0 +1,59 @@
...
+            }
+            return reject(Error(msg))
+          }
+  
+          if (status.isFinalized) resolve(status)
+        })
+    })
+  }
+
+  public async sendTransfer (toAddress: string, amount: string, startProgress?: ()=>void, stopProgress?: ()=>void) {
```
mixmix, line 44: This feels like `interaction` to me... I don't understand why it exists - if the progress things are optional then this wrapper only does
- formatting `amount`
- logging


I think logging should be in the base `transfer` anyway ...

mixmix (approved):
YUSSS. I'm gonna make minor changes then merge

src/sign/command.ts
mixmix: DONE

src/sign/command.ts
mixmix: done

src/transfer/main.ts
```diff
@@ -15,19 +15,17 @@ export class EntropyTransfer extends EntropyBase {
...
-      const transferStatus = await this.rawTransfer({
+      await this.rawTransfer({
         from: this.entropy.keyring.accounts.registration.pair,
         to: toAddress,
         amount: formattedAmount
       })
-      if (transferStatus.isFinalized) {
-        if (progress) return progress.stop()
-        return
-      }
```
mixmix: Dropped this check as `rawTransfer` only ever resolves if `isFinalized === true` already

tests/transfer.test.ts
```diff
@@ -25,61 +25,53 @@ test('Transfer', async (t) => {
...
+  const TransferService = new EntropyTransfer(charlieEntropy, endpoint)
+  const inputAmount = "1.5"
+  await run(
     'transfer',
-    TransferService.transfer({
-      from: charlieEntropy.keyring.accounts.registration.pair,
-      to: recipientAddress,
-      amount: BigInt(1000 * 10e10)
-    })
+    TransferService.transfer(naynayAddress, inputAmount)
```
mixmix, line 65: Testing the actual "main" function we're using with the input it will be receiving (e.g. `"1.5"` as a string)

src/transfer/main.ts
```diff
@@ -15,19 +15,17 @@ export class EntropyTransfer extends EntropyBase {
   // - progress callbacks (optional)
 
   async transfer (toAddress: string, amount: string, progress?: { start: ()=>void, stop: ()=>void }) {
-    const formattedAmount = BigInt(parseInt(amount) * 1e10)
+    const formattedAmount = BigInt(Number(amount) * 1e10)
```
mixmix, line 18: YO! the `parseInt` here was rounding our inputs.
Also what is our name for this `1e10`   ... is that that number of decimal places we support?

PR #220, opened 2024-08-28
signing-restructure tweeeeks

mixmix:


src/balance/interaction.ts
```diff
@@ -4,9 +4,9 @@ import { EntropyBalance } from "./main"
 export async function entropyBalance (entropy, endpoint, storedConfig) {
   try {
     const BalanceService = new EntropyBalance(entropy, endpoint)
-    const balanceString = await BalanceService.getAccountBalance(storedConfig.selectedAccount)
-    print(`Address ${storedConfig.selectedAccount} has a balance of: ${balanceString}`)
+    const balance = await BalanceService.getBalance(storedConfig.selectedAccount)
+    print(`Address ${storedConfig.selectedAccount} has a balance of: ${balance.toLocaleString('en-US')} BITS`)
```
rh0delta, line 8: Why was this change made?

mixmix: I killed `getAccountBalance` in `main.ts`
I noticed it was just `getBalance` with some print formatting .. which would only ever be used in `interaction.ts` (since `command.ts` wants pretty raw output, never e.g. `120 BITS` IMO)

src/balance/main.ts
```diff
@@ -11,29 +11,24 @@ export class EntropyBalance extends EntropyBase {
...
-  }
-
-  public async getAccountBalance (address: string) {
-    const balance = await this.getBalance(address)
 
-    this.logger.log(`Current balance of ${address}: ${balance}`, EntropyBalance.name)
-    
-    return `${balance.toLocaleString('en-US')} BITS`
+    return balanceInfo
   }
```
mixmix, line 33: All this function adds is formatting + logging - that seems more the domain of command/interaction, so have moved is there

src/sign/command.ts
```diff
@@ -24,22 +24,15 @@ function entropySign (entropy: Entropy, signCommand: Command) {
       )
     )
     .action(async (msg, opts) => {
-      if (opts.accountAddress) {
-        entropy = await reloadEntropy(
-          entropy,
-          opts.accountAddress,
-          entropy.keyring.accounts.registration.address, opts.endpoint
-        )
-      }
```
mixmix, line 33: The pre-hook covers this case already I think
- when startup happens `opts.account` is picked up (defaults to `config.selectedAccount`)
- `entropy` is started based on `opts.account`
- when a command finishes, the process is exited (i.e. `entropy` fails to exist as a variable)

src/sign/interaction.ts
```diff
@@ -1,13 +1,13 @@
...
 import inquirer from "inquirer"
 import Entropy from "@entropyxyz/sdk"
 import { EntropySign } from "./main"
 
 export async function entropySign (entropy: Entropy, endpoint: string) {
   const SigningService = new EntropySign(entropy, endpoint)
-  const { interactionChoice } = await inquirer.prompt(interactionChoiceQuestions)
-  switch (interactionChoice) {
+  // const { interactionChoice } = await inquirer.prompt(interactionChoiceQuestions)
+  // switch (interactionChoice) {
```
mixmix, line 10: no point asking the user what option they want when there is only 1

src/sign/utils.ts
```diff
@@ -51,29 +51,32 @@ export const rawSignParamsQuestions = [
...
-  const { messageAction } = await inquirer.prompt(messageActionQuestions)
-  switch (messageAction) {
-  case 'Text Input': {
-    const { userInput } = await inquirer.prompt(userInputQuestions)
-    msg = userInput
-    break
-  }
+  // let msg: string
+  // let msgPath: string
+  // const { messageAction } = await inquirer.prompt(messageActionQuestions)
```
mixmix, line 56: no point asking the user what option they want when there is only 1

src/common/masking.ts
```diff
@@ -7,6 +7,8 @@ const DEFAULT_MASKED_FIELDS = [
 ];
 
 export function maskPayload (payload: any): any {
+  if (typeof payload === 'string') return payload
```
mixmix, line 10: Was using the logs and noticed `cloneDeep(string) => object` and it was making super messed up logs...

mixmix:
I think I done fixed the test...

src/sign/command.ts
rh0delta: yeah, i should've removed this. previously there was an address argument and if that was supplied entropy needed to be reloaded. good catch tho

rh0delta (approved):

mixmix:
sweeet, thanks

PR #227, opened 2024-09-19
get all config.get* methods fail safely

mixmix:
I noticed that there was a nice error-handler for `getSync` but not for `get`!

Consolidated!

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -54,12 +54,11 @@ export async function init (configPath = CONFIG_PATH, oldConfigPath = OLD_CONFIG
     await set(newConfig, configPath)
   }
 }
-function noop () {}
 
 export async function get (configPath = CONFIG_PATH) {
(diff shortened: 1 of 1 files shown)


rh0delta (approved):

src/config/index.ts
```diff
@@ -54,28 +54,37 @@ export async function init (configPath = CONFIG_PATH, oldConfigPath = OLD_CONFIG
...
+
+/* util */
+function noop () {}
+
+function makeGetErrorHandler (configPath) {
+  return function getErrorHandler (err) {
+    if (err.code !== 'ENOENT') throw err
+
+    const newConfig = migrateData(allMigrations, {})
+    mkdirp.sync(dirname(configPath))
```
mixmix, line 86: caught a bug too!

YO we are gonna need to write some testing runsheets - the "fresh install" is a test-case we are missing a lot

PR #229, opened 2024-09-24
fix config (again)

mixmix:
So yeah the config was not good.... this chills it out, makes them config functions way simpler, safer.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -4,7 +4,8 @@
 import { Command, Option } from 'commander'
 
 import { EntropyTuiOptions } from './types'
-import { currentAccountAddressOption, endpointOption, loadEntropy } from './common/utils-cli'
+import { accountOption, endpointOption, loadEntropy } from './common/utils-cli'
+import * as config from './config'
 
 import launchTui from './tui'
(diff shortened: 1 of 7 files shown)


src/cli.ts
```diff
@@ -17,10 +18,9 @@ const program = new Command()
 /* no command */
 program
   .name('entropy')
-  .description('CLI interface for interacting with entropy.xyz. Running this binary without any commands or arguments starts a text-based interface.')
-  .addOption(currentAccountAddressOption())
+  .description('CLI interface for interacting with entropy.xyz. Running without commands starts an interactive ui')
```
rh0delta, line 21: This will need to be reverted, as the description being removed was written by @johnnymatthews and is the intended copy for the description of the cli command.

rh0delta (changes requested):

src/cli.ts
mixmix, line 21:
```suggestion
  .description('CLI interface for interacting with entropy.xyz. Running this binary without any commands or arguments starts a text-based interface.')
```

mixmix: DONE!

rh0delta (approved):

PR #242, opened 2024-09-26
refactor entropy register so CLI/TUI are similar

mixmix:
During QA discussion, notice that entropy register was coded the same in CLI + TUI.

This PR:
- reduces what `main` does to only be registering on-chain
- moves entropy config mutation + persistence into `util`
- changes `command` and `interaction` to use the same 2 functions
  - includes refactor of the `command` action

### DIFF
--- a/src/account/command.ts
+++ b/src/account/command.ts
@@ -1,11 +1,10 @@
 import Entropy from "@entropyxyz/sdk"
 import { Command, Option } from 'commander'
 import { EntropyAccount } from "./main";
-import { selectAndPersistNewAccount } from "./utils";
+import { selectAndPersistNewAccount, addVerifyingKeyToAccountAndSelect } from "./utils";
 import { ACCOUNTS_CONTENT } from './constants'
 import * as config from '../config'
 import { cliWrite, accountOption, endpointOption, loadEntropy, passwordOption } from "../common/utils-cli";
-import { findAccountByAddressOrName } from "src/common/utils";
 
 export function entropyAccountCommand () {
   return new Command('account')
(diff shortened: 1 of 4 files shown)


src/account/command.ts
```diff
@@ -102,28 +101,13 @@ function entropyAccountRegister () {
...
-      const { accounts } = storedConfig
-      const accountToRegister = findAccountByAddressOrName(accounts, account)
-      if (!accountToRegister) {
-        throw new Error('AccountError: Unable to register non-existent account')
-      }
-
-      const entropy: Entropy = await loadEntropy(accountToRegister.address, endpoint)
-      const accountService = new EntropyAccount(entropy, endpoint)
-      const updatedAccount = await accountService.registerAccount(accountToRegister)
+      // NOTE: loadEntropy throws if it can't find opts.account
```
mixmix, line 104:
```suggestion
      // NOTE: loadEntropy throws if it can't find opts.account
      // This is important because we don't want to register if we don't have a place to persist later
```

src/account/main.ts
```diff
@@ -75,28 +76,6 @@ export class EntropyAccount extends EntropyBase {
       })
   }
 
-  // WATCH: should this be extracted to interaction.ts?
-  async registerAccount (account: EntropyAccountConfig, registerParams?: AccountRegisterParams): Promise<EntropyAccountConfig> {
```
mixmix, line 79: All this function is is `registerAndPushKeysIntoAccount`

I have pulled this out into the verbosely named util which
- pushes into account
- persists config

src/account/utils.ts
```diff
@@ -19,11 +19,24 @@ export async function selectAndPersistNewAccount (newAccount) {
...
     ...storedConfig,
-    accounts,
     selectedAccount: newAccount.address
   })
 }
 
+export async function addVerifyingKeyToAccountAndSelect (verifyingKey: string, accountNameOrAddress: string) {
+  const storedConfig = await config.get()
+  const account = findAccountByAddressOrName(storedConfig.accounts, accountNameOrAddress)
+  if (!account) throw Error(`Unable to persist verifyingKey "${verifyingKey}" to unknown account "${accountNameOrAddress}"`)
```
mixmix, line 29: This is very unlikely because we have earlier checked the account exists...but we put a nice message out anyway :shrug: 

Alternative.. make a function with an uglier signature `(storeConfig, verifyingKey, account)` ? :cry:  this could still fail if `account` is not within `storedConfig.accounts` tho ...

src/account/utils.ts
```diff
@@ -19,11 +19,24 @@ export async function selectAndPersistNewAccount (newAccount) {
...
+
+  account.data.registration.verifyingKeys.push(verifyingKey)
+
+  // persist to config, set selectedAccount
+  await config.set({
+    ...storedConfig,
+    selectedAccount: account.address
+  })
```
mixmix, line 37: Check my working.

We don't need to any of the fancy slicing or inserting `accounts` here because:
- on line 28 : we grab the ref for target `account` in `storedConfig.accounts`
- on line 31 : we push a key into that account's data (mutating it)
- on line 35 : we say "persist that mutated storedConfig"

src/account/main.ts
frankiebee: it's also the lgger correct?

PR #250, opened 2024-10-03
fix cli opts bug with multiple accountOptions

mixmix:
tiny PRs!


src/cli.ts
```diff
@@ -20,8 +20,6 @@ const program = new Command()
 program
   .name('entropy')
   .description('CLI interface for interacting with entropy.xyz. Running this binary without any commands or arguments starts a text-based interface.')
-  .addOption(accountOption())
-  .addOption(endpointOption())
```
mixmix, line 24: THIS was the bug

src/common/utils-cli.ts
```diff
@@ -44,36 +44,43 @@ export function endpointOption () {
...
     .env('ENTROPY_ACCOUNT')
-    .argParser(async (account) => {
-      if (storedConfig && storedConfig.selectedAccount !== account) {
-        // Updated selected account in config with new address from this option
-        await config.set({
-          ...storedConfig,
-          selectedAccount: account
-        })
-      }
+    .argParser(addressOrName => {
```
mixmix, line 64: :bug:  - this cannot be async!

See comments

src/config/index.ts
```diff
@@ -73,14 +73,28 @@ export async function set (config: EntropyConfig, configPath = CONFIG_PATH) {
...
+  // no need for update
+
+  const newConfig = {
+    ...storedConfig,
+    selectedAccount: account.name
+  }
+  await set(newConfig, configPath)
+  return newConfig
+}
+
```
mixmix, line 89: :fire: one bug I saw was inconsistent setting of  `config.selectedAccount` as `account.name` OR `account.address`

I decided to fix it to be `name` because then in the `--help` you will see e.g  `(default: naynay)` instead of `(default: 0x12312231231232)` :skull: 

I then changed
- `accountOption` to deal with `name`
- added a special method `setSelectedAccount` to handle the persistence of the right thing

Things I'm not happy about:
1. repeated `config.get`...
2. this is all gonna be a pain when we want to add `-c, --config <configPath>`
3. `src/account/utils.js` still has hard-coded `config.set` for `selectedAccount` because... we need to persist the whole config anyway...

rh0delta:
@mixmix you able to update this PR and rebase off of the current version of dev? can complete my full review after that's done.

mixmix:
@rh0delta merged in `dev`  (not doing rebase because our merge/squash strategy makes rebasing a sad time)

rh0delta (approved):

PR #253, opened 2024-10-15
Mixmix/faucet tidy

mixmix:
I was having trouble with tests ... did a lot of debugging, tidying code, logging.
Finally realised it was my `/etc/hosts` file needed editing for this to work for faucet tests to pass

SO:
- documented to `/etc/hosts` file changes needed
- included all my tidying


.github/workflows/test.yaml
```diff
@@ -28,6 +28,6 @@ jobs:
         run: yarn run build
       - name: Add TSS server host mappings
         run: |
-          echo "127.0.0.1 alice-tss-server bob-tss-server" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 alice-tss-server bob-tss-server charlie-tss-server dave-tss-server" | sudo tee -a /etc/hosts
```
mixmix, line 31: we're gonna need this later

dev/bin/test-hosts.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# For local tests to work, we need to have aliases set up to point to 127.0.0.1
+# see: dev/README.md
+
```
mixmix, line 5: Save future devs time by testing the aliases are set up

src/common/constants.ts
```diff
@@ -0,0 +1 @@
+export const BITS_PER_TOKEN = 1e10
```
mixmix, line 1: extracted this so it's a constant

src/common/utils.ts
```diff
@@ -73,3 +73,21 @@ export function findAccountByAddressOrName (accounts: EntropyAccountConfig[], al
...
+    const { docs, name, section } = decoded
+
+    msg = `${section}.${name}: ${docs.join(' ')}`
+  } else {
+    // Other, CannotLookup, BadOrigin, no extra info
+    msg = dispatchError.toString()
+  }
+
+  return Error(msg)
+}
```
mixmix, line 93: I saw this code replicated across code base. 

TODO: rats I just noticed it uses "this" need to pass in `entropy`
```suggestion
export function formatDispatchError (entropy, dispatchError) {
  let msg: string
  if (dispatchError.isModule) {
    // for module errors, we have the section indexed, lookup
    const decoded = entropy.substrate.registry.findMetaError(
      dispatchError.asModule
    )
    const { docs, name, section } = decoded

    msg = `${section}.${name}: ${docs.join(' ')}`
  } else {
    // Other, CannotLookup, BadOrigin, no extra info
    msg = dispatchError.toString()
  }

  return Error(msg)
}
```

src/account/main.ts
```diff
@@ -83,20 +84,7 @@ export class EntropyAccount extends EntropyBase {
...
-                dispatchError.asModule
-              )
-              const { docs, name, section } = decoded
-  
-              msg = `${section}.${name}: ${docs.join(' ')}`
-            } else {
-              // Other, CannotLookup, BadOrigin, no extra info
-              msg = dispatchError.toString()
-            }
-            const error = Error(msg)
```
mixmix, line 99: I saw this code duplicated accross the code-base.
It's messy, it distracts from what's important, so I extracted it into a function in `common/utils`

mixmix:
Hmmm ... this was fine before. @rh0delta it looks like you've been writing code without reviewing / merging open PRs? I could be wrong. Anyway let's figure out how we mis-coordinated our work and fix that

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
 import { EntropyAccount } from '../src/account/main'
 
-async function setupAndFundFaucet (t, naynayEntropy) {
-  const { run, entropy, endpoint } = await setupTest(t, { seed: charlieStashSeed })
-  const accountService = new EntropyAccount(entropy, endpoint)
-  const transferService = new EntropyTransfer(entropy, endpoint)
-  const faucetService = new EntropyFaucet(naynayEntropy, endpoint)
+async function setupAndFundFaucet (t, naynay: Entropy) {
+  const { run, entropy: charlie, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+  // WARNING: setupTest is getting called twice ... which tries to spin up 2 networks? is this ok?
```
mixmix, line 15: :fire:

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
-  const faucetProgramPointer = await run('Deploy faucet program', entropy.programs.dev.deploy(faucetProgram, configurationSchema, auxDataSchema))
-  
+  const faucetProgramPointer = await run(
+    'Deploy faucet program',
+     charlie.programs.dev.deploy(faucetProgram, configurationSchema, auxDataSchema)
+)
+
   // Confirm faucetPointer matches deployed program pointer
   t.equal(faucetProgramPointer, LOCAL_PROGRAM_HASH, 'Program pointer matches')
+  // TODO: this is gonna be wrong
```
mixmix, line 49: This const should be pulled into the test, not over in src right?

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
+async function setupAndFundFaucet (t, naynay: Entropy) {
+  const { run, entropy: charlie, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+  // WARNING: setupTest is getting called twice ... which tries to spin up 2 networks? is this ok?
+  const charlieAddress = charlie.keyring.accounts.registration.address
 
-  const faucetProgram = readFileSync('tests/programs/faucet_program.wasm')
+  const account = new EntropyAccount(charlie, endpoint)
+  const transfer = new EntropyTransfer(charlie, endpoint)
+  // NOTE: naynay
+  const faucet = new EntropyFaucet(naynay, endpoint) // QUESTION: this seems like a mistage
```
mixmix, line 21: :fire: why would we be touching naynay at all in the faucet setup. This feels like a mistake to me?

tests/faucet.test.ts
rh0delta: setupTest was always getting called twice for the faucet tests since we needed two entropy instances, we already accepted this when the faucet was initially merged to dev

tests/faucet.test.ts
rh0delta: not sure what you mean by this, but the program has that is available in the constants is the same that is derived in the test, and will always be the same. The reason it differs from the current TEST_NET hash is because the faucet deployed to the testnet was without the auxiliary and configuration schemas. It needs to be redeployed

tests/faucet.test.ts
rh0delta: the entropy instance passed to the instantiation of the Faucet service must be the entropy account receiving the funds.

tests/faucet.test.ts
mixmix: Oh, I changes the schema to be real schema, which will result in a different hash. Just need to update

tests/faucet.test.ts
mixmix: Yes, but I think the way we want to set it up is Charlie sets up a faucet account?

src/transfer/main.ts
```diff
@@ -15,8 +18,7 @@ export class EntropyTransfer extends EntropyBase {
   // - progress callbacks (optional)
 
   async transfer (toAddress: string, amount: string, progress?: { start: ()=>void, stop: ()=>void }) {
-    const formattedAmount = BigInt(Number(amount) * 1e10)
-    // TODO: name this multiplier 1e10 somewhere
+    const formattedAmount = BigInt(Number(amount) * BITS_PER_TOKEN)
```
dvdplm, line 21: Would `DIGITS_PER_TOKEN` work here? Or something like "TOKEN_SCALING_FACTOR"?

mixmix: the naming was just based on the unit I'd heard others using - "Bit" and "Token".

- DIGITS .... feels like it could be more specific, but I'm not hard block
- SCALING_FACOTOR ... feels pretty generic / unclear. Scaling in what way?

I think what I should do is leave a good comment next to the definition. Will check that's done

src/common/constants.ts
mixmix, line 1:
```suggestion
/* 
  A "bit" is the smallest indivisible unit of account value we track.
  A "token" is the human readable unit of value value
  This constant is then "the number of bits that make up 1 token", or said differently
  "how many decimal places our token has".
*/
export const BITS_PER_TOKEN = 1e10
```

src/common/constants.ts
```diff
@@ -0,0 +1,7 @@
+/* 
+  A "bit" is the smallest indivisible unit of account value we track.
+  A "token" is the human readable unit of value value
+  This constant is then "the number of bits that make up 1 token", or said differently
+  "how many decimal places our token has".
+*/
```
mixmix, line 6: @dvdplm here's a bit more explainer. Happy for feedback :D

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
 import { EntropyAccount } from '../src/account/main'
 
-async function setupAndFundFaucet (t, naynayEntropy) {
-  const { run, entropy, endpoint } = await setupTest(t, { seed: charlieStashSeed })
-  const accountService = new EntropyAccount(entropy, endpoint)
-  const transferService = new EntropyTransfer(entropy, endpoint)
-  const faucetService = new EntropyFaucet(naynayEntropy, endpoint)
+async function setupAndFundFaucet (t) {
+  const { run, entropy: charlie, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+  // NOTE: there is some spin-up-network but it's idempotent!
```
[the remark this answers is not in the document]
mixmix:
```suggestion
  // NOTE: setupTest calles spinUpNetwork, but it's idempotent = chill!
```

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
 import { EntropyAccount } from '../src/account/main'
 
-async function setupAndFundFaucet (t, naynayEntropy) {
-  const { run, entropy, endpoint } = await setupTest(t, { seed: charlieStashSeed })
-  const accountService = new EntropyAccount(entropy, endpoint)
-  const transferService = new EntropyTransfer(entropy, endpoint)
-  const faucetService = new EntropyFaucet(naynayEntropy, endpoint)
+async function setupAndFundFaucet (t) {
+  const { run, entropy: charlie, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+  // NOTE: setupTest calles spinUpNetwork, but it's idempotent = chill!
```
mixmix, line 13:
```suggestion
  // NOTE: setupTest calls spinUpNetwork... which has already been called, but this is idempotent = chill!
```

tests/faucet.test.ts
```diff
@@ -1,94 +1,113 @@
...
 import { EntropyAccount } from '../src/account/main'
 
-async function setupAndFundFaucet (t, naynayEntropy) {
-  const { run, entropy, endpoint } = await setupTest(t, { seed: charlieStashSeed })
-  const accountService = new EntropyAccount(entropy, endpoint)
-  const transferService = new EntropyTransfer(entropy, endpoint)
-  const faucetService = new EntropyFaucet(naynayEntropy, endpoint)
+async function setupAndFundFaucet (t) {
+  const { run, entropy: charlie, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+  // NOTE: setupTest calls spinUpNetwork... which has already been called, but this is idempotent = chill!
```
mixmix, line 13: @dvdplm there last comment was terribly written! Here's a better comment.
Thanks for catching my end of day waffle

src/common/constants.ts
dvdplm: So to us "bits" is what "satoshi" is to bitcoin and "gwei" to ethereum? I may have opinions about the choice of such commonly used terms but beyond that: the comment is excellent, tyvm!

src/faucet/helpers/signer.ts
```diff
@@ -39,11 +39,11 @@ export default class FaucetSigner implements Signer {
...
       amount: this.amount
     }
-    
+
     const signature = await this.#entropy.sign({
       sigRequestHash: u8aToHex(raw),
       // @ts-ignore
-      hash: {custom: 0},
+      hash: { custom: 0 }, // TODO: hash should be a string? why is this an object?
       auxiliaryData: [auxData],
```
rh0delta, line 47: this is the custom hashing algo used for the faucet program. the program wasn't built to take a string as the hashing algorithm so had to update the hash property here to be of `any` type

rh0delta: it needs to stay as an object for now

mixmix:
@mixmix to freshen up then kick to naynay

src/faucet/helpers/signer.ts
mixmix: changed to NOTE

tests/faucet.test.ts
```diff
@@ -62,13 +74,16 @@ async function setupAndFundFaucet (t) {
 }
 
 test('Faucet Tests: Successfully send funds and register', async t => {
-  const { run, endpoint, entropy: naynay } = await setupTest(t)
   const { faucetAddress, chosenVerifyingKey, faucetProgramPointer } = await setupAndFundFaucet(t)
+  console.log(('faucet set up'))
```
rh0delta, line 78:
```suggestion
```

rh0delta (approved):

PR #261, opened 2024-10-18
[NayNay] Help text for command options

### LINKED ISSUE #244: 🔥  `entropy -a --help` does not help and nukes config
if you run the command above it will nuke your config. proceed with caution... and backups

rh0delta:
## Related Issue(s)

- Resolves #244 

## Proposed Changes

- contrary to intended use of the help option with commander, this PR adds handler when user adds help option after option that requires an argument
- for now, only updated account option, may require discussion with team if this is how we want to proceed

## Testing

- [ ] Manual testing performed

## Checklist

- [ ] I have performed a self-review of my code.
- [ ] I have added tests.

### DIFF
--- a/src/common/utils-cli.ts
+++ b/src/common/utils-cli.ts
@@ -1,6 +1,6 @@
 import Entropy from '@entropyxyz/sdk'
 import { Option } from 'commander'
-import { findAccountByAddressOrName, stringify } from './utils'
+import { findAccountByAddressOrName, print, stringify } from './utils'
 import * as config from '../config'
 import { initializeEntropy } from './initializeEntropy'
 
@@ -57,6 +57,16 @@ export function accountOption () {
     .argParser(addressOrName => {
       // We try to map addressOrName to an account we have stored
       if (!storedConfig) return addressOrName
+
+      const availableAccounts = {}
+      storedConfig.accounts.forEach(acct => availableAccounts[acct.name] = acct.address)
+      if (!Object.keys(availableAccounts).includes(addressOrName) || !Object.values(availableAccounts).includes(addressOrName)) {
+        print('=============AVAILABLE ACCOUNTS==============')
+        print(availableAccounts)
+        print('=============================================\n')
+        console.error(`AccountError: [${addressOrName}] is not a valid argument for the account option. Please use one of the aliases or addresses as shown above\n`)
+        process.exit(1)
+      }
 
       const account = findAccountByAddressOrName(storedConfig.accounts, addressOrName)
       if (!account) return addressOrName
--- a/src/common/utils-cli.ts
+++ b/src/common/utils-cli.ts
@@ -60,16 +60,15 @@ export function accountOption () {
 
       const availableAccounts = {}
       storedConfig.accounts.forEach(acct => availableAccounts[acct.name] = acct.address)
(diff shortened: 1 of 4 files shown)


mixmix:
@rh0delta says this needs more design work
Doesn't need to be in this release

Call on someone to discuss options e.g. Mix

mixmix:
discussed : should throw on `-a --help`

src/common/utils-cli.ts
```diff
@@ -58,6 +58,16 @@ export function accountOption () {
       // We try to map addressOrName to an account we have stored
       if (!storedConfig) return addressOrName
 
+      const availableAccounts = {}
+      storedConfig.accounts.forEach(acct => availableAccounts[acct.name] = acct.address)
+      if (!Object.keys(availableAccounts).includes(addressOrName) || !Object.values(availableAccounts).includes(addressOrName)) {
```
mixmix, line 63: :fire:  this is `addressOrName` so this check is not a good match.

`findAccountByAddressOrName` already handles the lookup gracefully. I think you wanna modify `if (!account) { }`

src/common/utils-cli.ts
```diff
@@ -58,6 +58,16 @@ export function accountOption () {
...
+        print('=============AVAILABLE ACCOUNTS==============')
+        print(availableAccounts)
+        print('=============================================\n')
+        console.error(`AccountError: [${addressOrName}] is not a valid argument for the account option. Please use one of the aliases or addresses as shown above\n`)
+        process.exit(1)
+      }
+
       const account = findAccountByAddressOrName(storedConfig.accounts, addressOrName)
       if (!account) return addressOrName
```
mixmix, line 72:
```suggestion
      if (!account) {
        console.error(`AccountError: [${addressOrName}] is not a valid argument for the account option.`)
        // if want to print accounts use some existing tool that formats them in the code?
        process.exit(1)
      }
```

rh0delta: updated to use this format, but kept the log of the account and addresses there

mixmix (approved):

mixmix:
Looks good to me - it meets the needs of the issue. We could print out more but your help text is plenty helpful IMO

src/common/utils-cli.ts
```diff
@@ -59,7 +59,11 @@ export function accountOption () {
       if (!storedConfig) return addressOrName
 
       const account = findAccountByAddressOrName(storedConfig.accounts, addressOrName)
-      if (!account) return addressOrName
+      if (!account) {
+        console.error(`AccountError: [${addressOrName}] is not a valid argument for the account option.`)
+        print(bold('!! Available accounts can be found using entropy account list || entropy account ls !!'))
+        process.exit(1)
```
mixmix, line 65: hmmm..  wonder why CI is breaking ...surely this doesn't need a `return` as well as `process.exit`?
The error in CI was "broken config"

mixmix: Could very well be a different issue / race

rh0delta: it's weird, it happens intermittently, but it happens on the program deploy method. might be a race condition with the tmp config file. don't think it needs to be completely solved now as it doesn't break cli

rh0delta: re-running the job results in a pass

PR #262, opened 2024-10-18
[NayNay] New SDK RC integration

rh0delta:
## Proposed Changes

- updated CLI with new sdk rc, and made changes to reflect the changes from core/sdk
- updated network type defualt to four-nodes
- added jump-start network to the initialization of entropy for both cli and tui
- updated all tests to reflect changes

## Testing

## Checklist

- [x] I have added tests.


tests/faucet.test.ts
```diff
@@ -8,11 +9,12 @@ import { EntropyFaucet } from '../src/faucet/main'
 import { LOCAL_PROGRAM_HASH } from '../src/faucet/utils'
 import { EntropyAccount } from '../src/account/main'
 
-async function setupAndFundFaucet (t, naynayEntropy) {
-  const { run, entropy, endpoint } = await setupTest(t, { seed: charlieStashSeed })
+async function setupAndFundFaucet (t) {
+  const { run, entropy, endpoint } = await setupTest(t, { seed: eveSeed })
```
rh0delta, line 13: @mixmix change to use eve entropy for faucet made here

src/common/utils-cli.ts
```diff
@@ -82,6 +82,9 @@ export async function loadEntropy (addressOrName: string, endpoint: string): Pro
   if (!selectedAccount) throw new Error(`No account with name or address: "${addressOrName}"`)
 
   const entropy = await initializeEntropy({ keyMaterial: selectedAccount.data, endpoint })
+  await entropy.substrate.tx.registry.jumpStartNetwork()
+    .signAndSend(entropy.keyring.accounts.registration.pair)
+
```
mixmix, line 87: :fire: This looks a lot like if you open the cli ... you send a signal to jump-start the network?!

I'm gonna go try read docs on what that does...

tests/account.test.ts
```diff
@@ -78,33 +80,50 @@ test('Account - import', async t => {
...
+  // the error found here was: RpcError: 1014: Priority is too low: (4471 vs 4471): The transaction has too low priority to replace another transaction already in the pool.
+  // which indicated a potential nonce issue, where the solution would be to manually wait for ~50 blocks to make any subsequent calls to the network. However,
+  // using naynay instance (fresh seed) jump start and the subsequent calls to the network (i.e register), the flow worked without issues.
+  const { entropy: naynay } = await setupTest(t, { networkType })
+
+  const accountService = new EntropyAccount(naynay, endpoint)
+  const transferService = new EntropyTransfer(entropy, endpoint)
+
+  // use of charlie seed is solely to transfer funds to the naynay (fresh seed) instance. these funds will be used for the jumpstart and register call
+  await run('transferring funds', transferService.transfer(naynay.keyring.accounts.registration.address, "1000"))
```
mixmix, line 98: I think the setup should be extracted (set up of a funded account).

mixmix: Done here https://github.com/entropyxyz/cli/pull/263

mixmix (changes requested):
see :fire: :point_up:

src/common/utils-cli.ts
rh0delta: yeah i agree this may not be the best place to trigger the jumpstart. need to find a better way to trigger the jumpstart when the network boots up

src/cli.ts
```diff
@@ -28,7 +28,7 @@ program
       .env('DEV_MODE')
       .hideHelp()
   )
-
+  .addOption(endpointOption())
```
mixmix, line 31: :fire: you cannot do this

There is a problem with Commander where if you put this here then when someone calls e.g. `entropy balance --endpoint url` then there are two options - this one, and the one attached at the root level `entropy --endpoint url`

and they will get smooshed and over-write each other. This was a bad bug I caught the other day.

This is so spicy I made an issue: https://github.com/entropyxyz/cli/issues/265

mixmix: MUST be removed for now

```suggestion
```

src/common/utils.ts
```diff
@@ -73,3 +73,10 @@ export function findAccountByAddressOrName (accounts: EntropyAccountConfig[], al
     accounts.find(account => account.name === aliasOrAddress)
   )
 }
+
+export async function jumpStartNetwork (entropy) {
+  await entropy.substrate.tx.registry.jumpStartNetwork()
+    .signAndSend(entropy.keyring.accounts.registration.pair)
```
mixmix, line 79: I am skeptical this promise actually waits for the jump-start to be done?
Should this be wrapped in `new Promise` with the event-listener which resolves the promise on `isFinalized` event or something?

src/tui.ts
```diff
@@ -129,8 +138,14 @@ async function main (entropy: Entropy, choices, options, logger: EntropyLogger)
         .catch(err => console.error('There was an error with program dev', err))
       break
     }
+    case 'Jump Start Network': {
+      await jumpStartNetwork(entropy)
+        .catch(err => console.error('There was an issue jumpstarting the network', err))
```
mixmix, line 143: I think this could use a a nice spinner thingee... if you felt like it

src/cli.ts
rh0delta: hmm, if that is the case then there is an even bigger issue at hand. If a user decides to use the TUI and wants to use a custom endpoint of their choosing they won't be able to. What do you think about creating a new endpoint option specific for this command? that way there is no overlap. We could also add some logic to determine which is the right option to use if there is a conflict between two of the same options attached to different commands?

src/common/utils-cli.ts
```diff
@@ -38,7 +38,32 @@ export function endpointOption () {
...
+export function tuiEndpointOption () {
+  return new Option(
+    '-et, --tui-endpoint <url>',
+    [
+      'Runs entropy with the given endpoint and ignores network endpoints in config.',
+      'Can also be given a stored endpoint name from config eg: `entropy --endpoint test-net`.'
+    ].join(' ')
+  )
+    .env('ENTROPY_ENDPOINT')
```
mixmix, line 54:
```suggestion
    .env('ENTROPY_TUI_ENDPOINT')
```

PR #264, opened 2024-10-22
mixmix/custom-config

mixmix:
This part of a PR chain:

[mixmix/config-types](https://github.com/entropyxyz/cli/pull/316) → [mixmix/load-entropy](https://github.com/entropyxyz/cli/pull/309) → **mixmix/custom-config**

---

The purpose of this PR is to be able to set a custom config file path for CLI

NOTES
- the `argParser` cannot be `async` 
  - => I think we're going to need to move config loading/parsing into the `.action`, probably `loadEntropy`
- the TUI cannot take the same `--config` as the other leaf actions
  - using the same option in multiple levels leads to collisions
  - if we really want this in TUI then we need to make the tui launch a subcommand like `entropy tui` or `entropy ui`


mixmix:
resolve conflicts, get ready for review

mixmix:
NOTE: this is passing all tests but will need a really close review + smoke test before merge.
We're gonna do this after this current release

mixmix:
@rh0delta I want you to try and break this. I think I got everything but dial up the pedantry with this one.

I turned of `config.set` taking a default config path, which then had typescript throwing everywhere it was not provided, but I bet there's still a place where something wrong has been written

src/common/utils-cli.ts
```diff
@@ -27,25 +30,10 @@ export function endpointOption () {
...
-      /* look up endpoint-alias */
-      const storedConfig = getConfigOrNull()
-      const endpoint = storedConfig?.endpoints?.[aliasOrEndpoint]
-      if (!endpoint) throw Error('unknown endpoint alias: ' + aliasOrEndpoint)
-
-      return endpoint
-    })
-    .default('wss://testnet.entropy.xyz/')
-    // NOTE: default cannot be "test-net" as argParser only runs if the -e/--endpoint flag
-    // or ENTROPY_ENDPOINT env set
```
rh0delta, line 43: what was the need to delete all of this? I'm all for it, just wondering the reason why

src/common/utils-cli.ts
```diff
@@ -54,30 +42,49 @@ export function accountOption () {
...
+export async function loadEntropy (opts: {
+  account: string,
+  config: string,
+  endpoint: string,
+}): Promise<Entropy> {
+  const storedConfig = getConfigOrNull(opts.config)
+  // NOTE: (mix) we expect config to be initialised (see hook in cli)
+  // ...there was some reason we wanted to preserve a `null` state,
+  // but I can't recall if it's still relevant, and we need to check
+  // the downstream ramifications of it being `null`
```
rh0delta, line 66: i think the null state was for fresh installs, where the config is created in the XDG standardized directory if there is no config found initially.

src/common/utils-cli.ts
```diff
@@ -54,30 +42,49 @@ export function accountOption () {
...
+  // if this account is not the default selectedAccount, make it so
+  if (storedConfig.selectedAccount !== account.name) {
+    await config.set(
+      {
+        ...storedConfig,
+        selectedAccount: account.name
+      },
+      opts.config
+    )
+  }
```
rh0delta, line 78: should we refactor this to be a separate function? might be cleaner

src/common/utils.ts
```diff
@@ -80,6 +83,17 @@ export function findAccountByAddressOrName (accounts: EntropyAccountConfig[], al
...
+export function absolutePath (somePath: string) {
+  switch (somePath.charAt(0)) {
+  case '.':
+    return join(process.cwd(), somePath)
+  case '~':
+    return join(homedir(), somePath.slice(1))
+  default:
+    return somePath
+  }
+}
```
rh0delta, line 95: 🥳

rh0delta (commented):
Everything looks good from reading the code, will take some time to test out everything locally tho, so will hold off on the approval

rh0delta (approved):

src/common/utils-cli.ts
mixmix: fair question!

The addition of the of custom-config meant that we cannot do the `argParser` stuff here, because to do that you need access to the `storedConfig`, but you don't know that without the `--config` loaded and parsed, which we cannot reach from in this function ...

SO I moved it all out into the `loadEntropy` domain. The `default` should perhaps still be in there tho :thinking:

src/common/utils-cli.ts
mixmix: there is a `setSelectedAccount` function, ... but we've not use that here because we already had `storedConfig` loaded... maybe we should change the signature of that function to be like 
```js
setSelectedAccount(configPath, account, { storedConfig? })
```

and then you load the config if `storedConfig` was not present. I was trying to save redundent disk reads + config parsing

mixmix:
e2e tests passing too : 
```bash
yarn test:network:up
./tests/e2e.cli.sh
yarn test:network:down
```

src/common/utils-cli.ts
mixmix: Oh this code is gone now :shrug:

Issue #265, opened 2024-10-23
TUI / CLI option collisions


mixmix:
We recently saw a problem where we wanted to be able to support options for both TUI and Programmatic CLI modes.
The problem is that if you e.g. `.addOption(endopintOption())` at the root and subcommand levels with... then the defaults and parsing of these options collide (see https://github.com/tj/commander.js/issues/2260)

As I see it, we're trending to one of several futures
1. the TUI only runs default config, OR << **we are currently here**
2. the TUI uses differently named options (e.g. `--tui-endpoint`), OR
3. the TUI is subcommand e.g. `entropy tui`, OR
4. the TUI is a different bin-script/ program e.g. `entropy-interactive`

rh0delta:
@mixmix since we have implemented option number 2, going to close this for now.

PR #270, opened 2024-10-30
move tui to a (default) subcommand

mixmix:
## Problem

I was hitting some problems with options colliding between the root `entropy` command and the other sub-commands like `entropy account register`.

## Solution

Reading the `commander` docs I saw there was an option for commands `{ default: true }`. This allows us to move tui functionality into a sub-command `entropy tui` AND sets this as the default command that is run (if no sub-command is called).

This resolves almost all the problems in https://github.com/entropyxyz/cli/issues/265 and means that we can have
- the same env for TUI and CLI
- no collision between the root command and subcommand options!

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5,7 +5,7 @@ import { Command } from 'commander'
 
 import * as config from './config'
 
-import { entropyTuiCommand as tui } from './tui'
+import { entropyTuiCommand as tui, tuiHelp } from './tui'
 import { entropyAccountCommand as account } from './account/command'
 import { entropyTransferCommand as transfer } from './transfer/command'
 import { entropySignCommand as sign } from './sign/command'
@@ -17,10 +17,7 @@ const cli = new Command()
 /* no command */
 cli
   .name('entropy')
-  .description([
-    'CLI interface for interacting with entropy.xyz.',
-    'Running this binary without any commands or arguments starts a text-based user interface (TUI).'
-  ].join(' '))
+  .description('CLI interface for interacting with entropy.xyz.')
 
   .addCommand(balance())
   .addCommand(account())
@@ -32,9 +29,20 @@ cli
   // The advantage of this is now the tui is a subcommand which means there
   // are no collisions with root program vs sub-command options
 
+  .addHelpText('before', [
+    tuiHelp(),
+    '',
+    '---',
+    '',
+  ].join('\n'))
+  .configureHelp({
+    // subcommandTerm (cmd) { return cmd.name() }, // show only name, not full signature
+  })
+
   .hook('preAction', async () => {
     // set up config file, run migrations
     return config.init()
   })
+
 
 cli.parseAsync()
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -4,13 +4,17 @@
 import { Command } from 'commander'
 
 import * as config from './config'
+import { print } from './common/utils'
 
-import { entropyTuiCommand as tui, tuiHelp } from './tui'
+import { entropyTuiCommand as tui, /* tuiAction */ } from './tui'
 import { entropyAccountCommand as account } from './account/command'
 import { entropyTransferCommand as transfer } from './transfer/command'
 import { entropySignCommand as sign } from './sign/command'
 import { entropyBalanceCommand as balance } from './balance/command'
 import { entropyProgramCommand as program } from './program/command'
+
+const packageVersion = 'v' + require('../package.json').version
+const coreVersion = process.env.ENTROPY_CORE_VERSION.split('-')[1]
 
 const cli = new Command()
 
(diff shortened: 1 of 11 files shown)


mixmix:
In testing this, I discovered that `entropy --help` was outputting help for all the commands (good) but that there was no mention of `--version`. I've added a commit which gives you output about the two major use paths:

```ouput
$ entropy --help

Usage: entropy [options]

TUI (text user interface) for interacting with entropy.xyz. A good place to start.

Options:
  -v, --version                 Displays the current running version of Entropy CLI
  -cv, --core-version           Displays the current running version of the Entropy Protocol
  -a, --account <name|address>  Sets the account for the session. Defaults to the last set account (or the first account if one has not been set before). (default: "mix", env: ENTROPY_ACCOUNT)
  -e, --endpoint <url>          Runs entropy with the given endpoint and ignores network endpoints in config. Can also be given a stored endpoint name from config eg: `entropy --endpoint test-net`. (default: "wss://testnet.entropy.xyz/", env:
                                ENTROPY_ENDPOINT)
  -h, --help                    display help for command


---

Usage: entropy [options] [command]

CLI interface for interacting with entropy.xyz.

Options:
  -h, --help                                 display help for command

Commands:
  balance [options] <account>                Command to retrieive the balance of an account on the Entropy Network
  account                                    Commands to work with accounts on the Entropy Network
  transfer [options] <destination> <amount>  Transfer funds between two Entropy accounts.
  sign [options] <msg>                       Sign a message using the Entropy network. Output is a JSON { verifyingKey, signature }
  program                                    Commands for working with programs deployed to the Entropy Network
  tui [options]                              TUI (text user interface) for interacting with entropy.xyz. A good place to start.
  help [command]                             display help for command
```

mixmix:
I think I want to discuss this as a team. 

## Requirements

Any good solution will satisfy

1. `entropy --help` is clear
    - shows tui options (if `entropy [options]` is the tui command)
    - shows commands
2. `entropy -v` and `entropy [command] -vk <string>` coexist
3. `entropy -e ws://testnet.entropy.xyz` and `entropy [command] -e `-e ws://testnet.entropy.xyz` coexist

## Option A - split out `entropy-tui`

Solves both (1) + (2) + (3) trivially. We can even mention there is a tui option in the `entropy --help`

## Option B - tui as default sub-command

This is what this PR is doing.
:cry:  (1) semi-fails ... I got into the source code of `commander` to hack the help output ... it's possible to do maybe better than what I have but it gets ugly fast.
:heavy_check_mark: (2) succeeds
:heavy_check_mark: (3) succeeds

## Option C - tui has root command

This is how `dev` currently is. It sucks because we have to use `ENTROPY_TUI_ENDPOINT` which we have already seen needlessly trip up Peg.

:heavy_check_mark: (1) yes
:x: (2) fail
:x: (3) fail

For (3) We *could* get very fancy and check `program.opts` versus `command.opts` and make sure the opts get through to the right level of action .. and not-overridden by a default.... there's a positional opts config we could mess with but then that makes the signatures VERY rigid.

## Option D - `entropy tui`

similar to (B), except `entropy` alone now does nothing.
We can recommend that a user runs `entropy tui` if they are new

---

I think we should do A or D

mixmix:
## Option E - no flag collisions

Replace `-vk` with e.g. `-k`
Probably replace `-pmk` with `-p`

The whole short-code is designed to be a single letter. We're hitting `-v` vs `-vk` tension exactly because of how short-codes work (`-vk` = `-v` + `-k`)

I strongly believe we should NOT change `-v, --version`

mixmix:
Ok! where this has settled on:

- `entropy` now just prints help
```output
Usage: entropy [options] [command]

CLI interface for interacting with entropy.xyz.

Options:
  -v, --version                              Displays the current running version of Entropy CLI
  -cv, --core-version                        Displays the current running version of the Entropy Protocol
  -h, --help                                 display help for command

Commands:
  tui [options]                              Text-based User Interface (interactive)
  account                                    Commands to work with accounts on the Entropy Network
  sign [options] <msg>                       Sign a message using the Entropy network. Output is a JSON {
                                             verifyingKey, signature }
  balance [options] <account>                Command to retrieive the balance of an account on the Entropy
                                             Network
  transfer [options] <destination> <amount>  Transfer funds between two Entropy accounts.
  program                                    Commands for working with programs deployed to the Entropy
                                             Network
```

- `entropy tui` launches the tui
- `entropy tui --help` gives help only for the tui
```output
Usage: entropy tui [options]

Text-based User Interface (interactive)

Options:
  -a, --account <name|address>  Sets the account for the session. Defaults to the last set account (or the
                                first account if one has not been set before). (default: "mix", env:
                                ENTROPY_ACCOUNT)
  -e, --endpoint <url>          Runs entropy with the given endpoint and ignores network endpoints in config.
                                Can also be given a stored endpoint name from config eg: `entropy --endpoint
                                test-net`. (default: "wss://testnet.entropy.xyz/", env: ENTROPY_ENDPOINT)
  -h, --help                    display help for command
```

There are now no collisions between a "root command" and "sub-command" on:
- endpoint (`-e` and `ENTROPY_ENDPOINT` work same for TUI/ CLI)
- account (this was broken for tui!)

src/cli.ts
```diff
@@ -1,61 +1,58 @@
...
+    }
+    if (opts.coreVersion) {
       print(coreVersion)
       process.exit(0)
     }
-    // NOTE: on initial startup you have no account
-    launchTui(entropy, opts)
+
+    // print entropy help and exit
+    cli.help()
```
rh0delta, line 46: is displaying the help text when a user types only `entropy` what we expect to happen? Also should we update the package json script for `yarn start` to start the TUI instead of just displaying the help text?

src/tui.ts
```diff
@@ -59,7 +73,18 @@ export default function tui (entropy: Entropy, options: EntropyTuiOptions) {
...
+async function setupConfig () {
+  let storedConfig = await config.get()
+
+  // set selectedAccount if we can
+  if (!storedConfig.selectedAccount && storedConfig.accounts.length) {
+    storedConfig = await config.setSelectedAccount(storedConfig.accounts[0])
+  }
+
+  return storedConfig
+}
```
rh0delta, line 87: just a thought: do we need to implement this for the CLI too? or are we assuming that a user will have an account selected when using the programmatic cli?

rh0delta:
@mixmix do we want to include this in the current release?

rh0delta (approved):

mixmix:
Good question. I don't see why not?
I think th just needs breaking change in changelog then merge

mixmix:
Needs CHANGELOG

mixmix:
cc @johnnymatthews breaking change to CLI / TUI

PR #288, opened 2024-11-07
animated logo *swipe*

mixmix:
boom


rh0delta (approved):

rh0delta:
if this is something new, i think we should still merge it to dev and update the changelog with a new unreleased version.

rh0delta:
want to add a changelog entry too?

PR #309, opened 2024-11-14
mixmix/load-entropy

mixmix:
This is part of a PR chain:

[mixmix/config-types](https://github.com/entropyxyz/cli/pull/316) → **mixmix/load-entropy** → [mixmix/custom-config](https://github.com/entropyxyz/cli/pull/264)

---

This a precursor PR for the custom-config work.

In noticed that we needed some guarantees about error handling, and that the current setup for how entropy gets loaded makes that really hard to manage / predict because of leftover messy code.

This PR
- extracts `loadEntropy` into a function
- kills `initializeEntropy`
- refactors the code into bite-sized functions
  - collect related concerns into a nice blocks of code
  - reduce the redundant checking of attributes everywhere


src/config/types.ts
```diff
@@ -1,5 +1,5 @@
 export interface EntropyConfig {
-  accounts: EntropyAccountConfig[]
+  accounts: EntropyConfigAccount[]
```
mixmix, line 2: Oh yeah, I renamed the types so they are more consistent / predictable
data | type
---|---
`storedConfig` | `EntropyConfig`
`storedConfig.accounts[0]` |  `EntropyConfigAccount`
`storedConfig.accounts[0].data` | `EntropyConfigAccountData`
`prettyPrint(storedConfig.accounts[0])` | `EntropyConfigAccountFormatted`

mixmix:
Ah ... this new `loadEntropy` is too tightly coupled to persistence for current testing....
Working on that now

src/common/load-entropy.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+interface ConfigGetterSetter {
+  get: () => Promise<EntropyConfig>
+  set: (config: EntropyConfig) => Promise<void>
+}
+
+// This expects a config getter/setter, and just throws errors it hits
+// Higher level functions are expected to define getter/setter and error handling
+// DO NOT EXPORT
+async function loadEntropy (opts: LoadEntropyOpts): Promise<Entropy> {
```
mixmix, line 113: This file has a base function `loadEntropy` which is not exported. What's exported are 3 helper functions
- `loadEntropyCli`
- `loadEntropyTui`
- `loadEntropyTest`

This allows us to customise the error-handling for each use-case.

I also extracted `config` as a dependency, so now there is just a `configGetterSetter` expected. This was my approach to making tests not depend on file-system, and allowing me to control the setup of config in that case

src/common/load-entropy.ts
```diff
@@ -0,0 +1,259 @@
...
+      async set (config: EntropyConfig) {
+        return fsConfig.set(config, opts.config)
+      }
+    }
+  })
+    .catch(err => {
+      console.error(`Loading Entropy account "${opts.account}" failed:`)
+      console.error(err)
+      // NOTE: deliberately does not exit process
+      // TODO: decide how to handle error...
```
mixmix, line 58: Not sure how `loadEntropy` errors should be handled in the TUI.... it probably means something bad with config was found, or the account you tried to load was bung in some way.... 

Question: this change gives us the tools to decide later ... do we need to decide now, or should we just merge and see?

src/account/utils.ts
```diff
@@ -28,7 +28,7 @@ export async function selectAndPersistNewAccount (newAccount: EntropyAccountConf
   })
 }
 
-export async function addVerifyingKeyToAccountAndSelect (verifyingKey: string, accountNameOrAddress: string) {
+export async function persistVerifyingKeyToAccount (verifyingKey: string, accountNameOrAddress: string) {
```
mixmix, line 31: I changed this function because
- it was doing 2 things
- the `selectedAccount` is already handled by `loadEntropy`

src/account/command.ts
```diff
@@ -106,12 +107,11 @@ function entropyAccountRegister () {
...
     // )
     .action(async (opts) => {
-      // NOTE: loadEntropy throws if it can't find opts.account
-      const entropy: Entropy = await loadEntropy(opts.account, opts.endpoint)
+      const entropy: Entropy = await loadEntropyCli(opts)
       const accountService = new EntropyAccount(entropy, opts.endpoint)
 
       const verifyingKey = await accountService.register()
-      await addVerifyingKeyToAccountAndSelect(verifyingKey, opts.account)
+      await persistVerifyingKeyToAccount(verifyingKey, opts.account)
```
mixmix, line 114: Because we have this, we don't need keyring listening to and trying to persist changes.

src/account/interaction.ts
```diff
@@ -87,7 +87,7 @@ export async function entropyRegister (entropy: Entropy, endpoint: string, store
   print("Attempting to register the address:", account.address)
   try {
     const verifyingKey = await accountService.register()
-    await addVerifyingKeyToAccountAndSelect(verifyingKey, account.address)
+    await persistVerifyingKeyToAccount(verifyingKey, account.address)
```
mixmix, line 90: Because we have this, we don't need keyring listening to and trying to persist changes.

(the only `account-update` emits are on `verifyingKey` changing (I hope!)

src/config/types.ts
mixmix: Moved this out into https://github.com/entropyxyz/cli/pull/316

src/account/utils.ts
mixmix: Here's where the emit is being stimulated in SDK

https://github.com/entropyxyz/sdk/blob/36ba894a95da3ad689ede5e2b4ba227034294b2d/src/index.ts#L147-L149

rh0delta (approved):
Tested out these changes locally, everything looks to be working as expected

PR #310, opened 2024-11-14
[NayNay] Balance --all

### LINKED ISSUE #305: entropy balance --all
Request: `entropy balance` should have a mode (maybe `--all`) where it outputs the balances of all admin accounts

Mix: I think this is a nice-to have/ usability

rh0delta:
## Related Issue(s)

- Resolves #305

## Proposed Changes

- updated balance command with an option to view all balances of accounts in the config

## Testing

## Screenshots (if applicable)

Screenshot 2024-11-14 at 2 40 39 PM

## Checklist

- [ ] I have added tests.


rh0delta:
this will need updating if #306 is merged before this PR

src/balance/command.ts
```diff
@@ -5,25 +5,40 @@ import { EntropyBalance } from "./main";
...
+import { formattedBalances } from "./utils";
 
 export function entropyBalanceCommand () {
   const balanceCommand = new Command('balance')
+  // account can no longer be a required argument if a user wishes to
+  // view the balances of all accounts
   balanceCommand
     .description('Command to retrieive the balance of an account on the Entropy Network')
-    .argument('<account>', [
+    .argument('[account]', [
```
mixmix, line 17: :fire: this changes the `account` argument to "optional" ... but it's only optional in the case that `--all` was passed.
This matters because `commander` normally does some work for us with helpful errors if a user does not supply a `<required>` argument. We can go off-piste, but then we need to build up our own guard rails

src/balance/command.ts
```diff
@@ -5,25 +5,40 @@ import { EntropyBalance } from "./main";
...
       'The address an account address whose balance you want to query.',
       'Can also be the human-readable name of one of your accounts'
     ].join(' '))
     .addOption(endpointOption())
+    .option('--all', 'Get balances for all admin accounts in the config')
     .action(async (account, opts) => {
```
mixmix, line 23: I don't like this:
- `account` used to be able to be an `address` OR `alias` for an account, you have broken that with `tempAddress`
- there should be only 2 paths through this but you have accidentally made 3:
  - :heavy_check_mark: `entropy balance naynay`  or `entropy balance 123adsd12312312asads987398172bb`
  - :heavy_check_mark: `entropy balance --all`
  - :x: `entropy balance` - this should error. It does not here yet.

Here's how I'd code it

```js
.action(async (account, opts) => {
  if (account) {
    // use old code?
  }
  else if (opts.all) {
    // some new code
  }
  else {
    // write error
    process.exit(1)
  }
})
```

maybe once I'd seen that, then look for ways to refactor

src/balance/utils.ts
```diff
@@ -1 +1,9 @@
-export const hexToBigInt = (hexString: string) => BigInt(hexString)
\ No newline at end of file
+import { BalanceInfo } from "./types"
+
+export const hexToBigInt = (hexString: string) => BigInt(hexString)
+
+export function formattedBalances (balanceInfo: BalanceInfo) {
```
mixmix, line 5: two questions:
1. we need to decide if we're outputting units on the end of this or not. I'm pretty keen on not. Can be another PR though after we decide finally across the repo.
2. Can this please be an Array?
  It seems silly maybe, but it makes for easier parsing in other scripts (notice how you had to write `Object.keys` for example). It also makes it easier to add more info later, which isn't so easy if you lock in ` { [address]: balance }`

Example solution: 
```js
[
  { address: 'aaaaaaaaaaaaaaaaaaaaaaa', bits: 1.12345688 },
  { address: 'bbbbbbbbbbbbbbbbbbbbbbb', bits: 27.2000013 }
]
```

mixmix: re: (2) maybe this is a different issue and we should have a look what we do elsewhere. 

`entropy account ls` uses Array

mixmix (changes requested):
Changes requested. Looking at the issue that generated this, I see it was under-specified. My/ our bad for having something in up-next like that. 

I think in those cases where we haven't shown expected input/ output, we should write an example first and review that. It's easier/ quicker than building something with no design.

src/balance/command.ts
rh0delta: updated in newest commit

src/balance/command.ts
rh0delta: added pseudo-error handling, whereas instead displaying an error, the help text is displayed

mixmix:
I'll have a go resolving these

src/balance/command.ts
```diff
@@ -2,30 +2,65 @@ import { Command } from "commander";
...
+      let entropy: Entropy
+      if (!account && opts.all) {
+        const tempAddress = accounts[0].address
+        entropy = await loadEntropy(tempAddress, opts.endpoint)
+      } else if (account) {
+        entropy = await loadEntropy(account, opts.endpoint)
+      } else {
+        balanceCommand.help()
+        return process.exit(0)
+      }
```
mixmix, line 35: :fire: this still has a gap: 
```bash
entropy balance naynay -a
```
`!account && opts.all` => `false`
`account` => `true`  but I specified `-a` as well... should not be allowed

So perhaps
```suggestion
      if (!account && opts.all) {
        const tempAddress = accounts[0].address
        entropy = await loadEntropy(tempAddress, opts.endpoint)
      } else if (account && !opts.all) {
        entropy = await loadEntropy(account, opts.endpoint)
      } else {
        balanceCommand.help()
        return process.exit(0)
      }
```

src/balance/command.ts
```diff
@@ -2,30 +2,65 @@ import { Command } from "commander";
...
+            return infos.map(info => {
+              return {
+                account: info.address,
+                balance: info.balance !== undefined
+                  ? toBits(info.balance)
+                  : info.error,
+                symbol
+              }
+            })
+          })
```
mixmix, line 55: I refactored this:
- `getBalances` returns an Array => everything is easier to map
- have `balances` output roughly match the `balance` output
   - `[{ account, balance, symbol }]`
   - `{ balance, symbol }`

Perhaps the singular should include the `account` too?

mixmix (changes requested):
Hey @rh0delta I merged `dev` into this and tweaked functions as I went. This is a sign I'm tired (in merging I was trying to understand the code, so I moved it around to make sure I understood it ) ... not ideal :hankey: sorry

mixmix:
more merge conflicts @rh0delta

mixmix (approved):

mixmix:
I can resolve that

src/balance/command.ts
```diff
@@ -2,32 +2,68 @@ import { Command } from "commander";
...
     .description('Command to retrieive the balance of an account on the Entropy Network')
-    .argument('account <address|name>', [
+    .argument('[account] <address|name>', [
       'The address an account address whose balance you want to query.',
       'Can also be the human-readable name of one of your accounts'
     ].join(' '))
+    .option('-a, --all', 'Get balances for all admin accounts in the config')
     .addOption(configOption())
     .addOption(endpointOption())
+    .addOption(configOption())
```
rh0delta, line 25:
```suggestion
```

PR #314, opened 2024-11-21
Mixmix/refactor scripts

mixmix:
**Problems**:
- :fire: `yarn test` doesn't currently reset the network before running tests => failures because tests assume unique program hashes (that have not been deployed already)
- :cry: the `package.json` script `yarn test` is a mess

**I've refactored the tests to:**
- [x] collected all the "ts" related test-steps into `test:ts:full`
- [x] separate out the "global" testing into `test:global`
  - rename: `test/global.test.ts` => `test/global.test.mjs` (no need to depend on building `tsx`)
  - required making some `testing-utils` into `.mjs` files
- [x] added `network:reset` script
- [x] moved `tap-spec` to be in the `package.json` only, making it easier to rip out / turn off / swap out / maintain

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,19 +9,22 @@
     "build": "./dev/bin/build.sh",
     "build:global": "yarn build && npm install -g",
     "lint": "eslint . --ext .ts --fix",
-    "test": "yarn test:types && yarn build:global && yarn test:hosts && yarn test:network:up && yarn test:ts && yarn test:network:down && yarn test:only",
+    "test": "yarn test:types && yarn test:global && yarn test:ts:full && yarn test:only",
+    "test:global": "yarn build:global && node tests/global.test.mjs | tap-spec",
     "test:hosts": "./dev/bin/test-hosts.sh",
     "test:only": "./dev/bin/test-only.sh",
-    "test:ts": "./dev/bin/test-ts.sh",
+    "test:ts": "./dev/bin/test-ts.sh | tap-spec",
+    "test:ts:full": "yarn test:hosts && yarn test:network:reset && yarn test:ts && yarn test:network:down",
     "test:types": "tsc --project tsconfig.json",
     "test:network:up": "node tests/testing-utils/test-network.mjs up",
     "test:network:down": "node tests/testing-utils/test-network.mjs down",
+    "test:network:reset": "yarn test:network:down && yarn test:network:up",
     "prepare": "husky",
     "prepack": "pinst --disable",
     "postpack": "pinst --enable",
-    "link:sdk": "yarn link @entropyxyz/sdk",
-    "unlink:sdk": "yarn unlink @entropyxyz/sdk",
-    "re-link:sdk": "yarn unlink:sdk && yarn link:sdk"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -9,19 +9,22 @@
...
     "build:global": "yarn build && npm install -g",
     "lint": "eslint . --ext .ts --fix",
-    "test": "yarn test:types && yarn build:global && yarn test:hosts && yarn test:network:up && yarn test:ts && yarn test:network:down && yarn test:only",
+    "test": "yarn test:types && yarn test:global && yarn test:ts:full && yarn test:only",
+    "test:global": "yarn build:global && node tests/global.test.mjs | tap-spec",
     "test:hosts": "./dev/bin/test-hosts.sh",
     "test:only": "./dev/bin/test-only.sh",
-    "test:ts": "./dev/bin/test-ts.sh",
+    "test:ts": "./dev/bin/test-ts.sh | tap-spec",
+    "test:ts:full": "yarn test:hosts && yarn test:network:reset && yarn test:ts && yarn test:network:down",
```
mixmix, line 17: `test:ts:full` bundles all the things needed to run the `ts` tests, saves space on `test`

package.json
```diff
@@ -9,19 +9,22 @@
...
+    "test:network:reset": "yarn test:network:down && yarn test:network:up",
     "prepare": "husky",
     "prepack": "pinst --disable",
     "postpack": "pinst --enable",
-    "link:sdk": "yarn link @entropyxyz/sdk",
-    "unlink:sdk": "yarn unlink @entropyxyz/sdk",
-    "re-link:sdk": "yarn unlink:sdk && yarn link:sdk"
+    "sdk:link": "yarn link @entropyxyz/sdk",
+    "sdk:unlink": "yarn unlink @entropyxyz/sdk",
+    "sdk:relink": "yarn sdk:unlink && yarn sdk:link"
```
mixmix, line 27: oh yeah, renamed these to follow the pattern of `<domain>:<sub-domain>` seen elsewhere in scripts

tests/testing-utils/setup-test.ts
```diff
@@ -53,3 +54,7 @@ export async function setupTest (t: Test, opts?: SetupTestOpts): Promise<{ entro
 
   return { entropy, run, endpoint }
 }
+
+function makeSeed () {
+  return '0x' + randomBytes(32).toString('hex')
+}
```
rh0delta, line 60: There's a function we use from polkadot's crypto utils to generate the seed, can we do the same here as well?

Re: https://github.com/entropyxyz/cli/blob/main/src/account/main.ts#L18

rh0delta (changes requested):

tests/testing-utils/setup-test.ts
mixmix: this is just a refactor, but I'll see if I can squeeze that in to!

mixmix:
Cool, passing locally with that change. I'll merge after CI passes

rh0delta (approved):

PR #315, opened 2024-11-22
[NayNay] Balance for any address & transfer (using only substrate api)

### LINKED ISSUE #237: `entropy balance <address>` should take any address
right now you have to have the address in your config but that command should take any address.

rh0delta:
## Related Issue(s)

- Resolves #237 

## Proposed Changes

- updated balance for prg cli to be able to retrieve balance for any address
- using newly exposed utility method to getSubstrate from sdk

## Testing

## Checklist

- [x] I have added tests.


rh0delta:
Build's will continue to fail for this PR until the addition to the sdk is released.

Re: https://github.com/entropyxyz/sdk/pull/435/files

mixmix:
@mixmix reviewing this
@frankiebee working on SDK thing for this

package.json
```diff
@@ -48,7 +48,7 @@
   },
   "homepage": "https://github.com/entropyxyz/cli#readme",
   "dependencies": {
-    "@entropyxyz/sdk": "0.4.0",
+    "@entropyxyz/sdk": "0.4.1-0",
```
mixmix, line 51: **TODO**: wait for release before merge

tests/balance.test.ts
```diff
@@ -50,3 +51,17 @@ test('getBalance + getBalances', async (t) => {
...
+  const run = promiseRunner(t)
+  // create new account, not saved in config
+  const newAccount = await EntropyAccount.create({ name: 'TestAnyBalance1' })
+
+  const balance = await run(
+    'getAnyBalance (newAccount)',
+    EntropyBalance.getAnyBalance(DEFAULT_ENDPOINT, newAccount.address)
+  )
+  t.equal(balance, 0, 'newSeed balance = 0')
+
```
mixmix, line 65: :fire: I'd like to see a test where you use this service to check the balance of e.g. eve and show it's large

mixmix: :fire: I'd like to see a test for 

```js
EntropyBalance.getAnyBalance(DEFAULT_ENDPOINT, "poop")
```

src/balance/command.ts
```diff
@@ -1,28 +1,16 @@
...
+    .argument('<address>', 'Any SS58 address you would like to get the balance of')
     .addOption(endpointOption())
-    .action(async (account, opts) => {
-      const entropy: Entropy = await loadEntropy(account, opts.endpoint)
-      const BalanceService = new EntropyBalance(entropy, opts.endpoint)
-
-      const { accounts } = await config.get()
-      const address = findAccountByAddressOrName(accounts, account)?.address
-
-      const balance = await BalanceService.getBalance(address)
```
mixmix, line 25: :fire: no longer supports `entropy balance naynay` - i.e. a breaking change to the API

I'd like to see that supported, and the argument called `<account>` still

mixmix (changes requested):
Excited for this. Needs more tests to prove it works, and needs to not regress the API

src/balance/command.ts
rh0delta: updated balance command to still support config accounts, and changed argument back to account

tests/balance.test.ts
rh0delta: add tests

mixmix:
merge conflicts now here too @rh0delta

mixmix:
Found some bugs that I fixed in this PR:  https://github.com/entropyxyz/cli/pull/327

mixmix:
Found that when I have a newly fauceted account this function doesn't work:

```bash
➜  cli git:(mixmix/any-balance) entropy balance --all | jq
[
  {
    "account": "5HNMFB1fELqSAEMkRtUMA6ACs1jD693h16Gt8ixvBS6csJRq",
    "balance": 2,
    "symbol": "BITS"
  },
  {
    "account": "5E4DpqmtnxBA6WbExNvFnJv2Hw5BsUpk8KnAVZLbBMWZVTXb",
    "balance": 0,
    "symbol": "BITS"
  }
]
➜  cli git:(mixmix/any-balance) entropy balance 5HNMFB1fELqSAEMkRtUMA6ACs1jD693h16Gt8ixvBS6csJRq | jq
Provided [account=5HNMFB1fELqSAEMkRtUMA6ACs1jD693h16Gt8ixvBS6csJRq] is not a valid substrate address
```

mixmix:
yeah have tried with a registered account and still getting that error "not a valid substrate address"

mixmix:
ok fixed it in that PR : https://github.com/entropyxyz/cli/pull/327/commits/619ff78c3392327f17d323011bd302710b1e26e4

@rh0delta I've seen us slip with if/ else + boolean logic a couple times now. I didn't spot it either till I did some closer testing. I think we need to EITHER write tests over such things, OR try to write much simpler (and more verbose) if/ else if/ else logic. Have a look at the solution in that linked commit. It's more nesting, but more clear (to me at least)

mixmix:
Mix to re-review

src/common/substrate-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+export async function getLoadedSubstrate (endpoint: string) {
+  const substrate = createSubstrate(endpoint)
+  await substrate.isReadyOrError
+  return substrate
+}
+
+export async function closeSubstrate (substrate: any) {
+  // closing substrate
+  return await substrate.disconnect()
+    .catch(err => console.error('Error closing connection', err.message))
```
mixmix, line 13: :fire: I think this should log AND throw ?? like if you don't throw this here swallows the error ... and then what? I think it could cause problems if you expect something to have done it's job and it just doesn't?

src/balance/main.ts
```diff
@@ -1,26 +1,20 @@
...
 import * as BalanceUtils from "./utils"
 import { BalanceInfo } from "./types"
 
-const FLOW_CONTEXT = 'ENTROPY-BALANCE'
-export class EntropyBalance extends EntropyBase {
-  constructor (entropy: Entropy, endpoint: string) {
-    super({ entropy, endpoint, flowContext: FLOW_CONTEXT })
-  }
+export class EntropyBalance {
+  constructor () {}
```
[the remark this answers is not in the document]
mixmix: perhaps like `EntropyBase` but call it `SubstrateBase`

mixmix (changes requested):
I like this, it's really interesting seeing how little we might need the full SDK... also begs the question "what is it even used for", "what is our policy for when we use SDK and when we use raw substrate?"

I'd like to see us not go to full class methods - I think keeping some of the pattern of `EntropyBase` but making a `SubstrateBase` class that we inherit from will keep the patterns of the repo more coherent and give use support for the logging and other things that we need, tidy up need to pass around `substrate` etc.

Have left more notes and thinking as I went with more detail.
Nice work!

package.json
rh0delta: it has not, i believe the full release will come when the signatures stuff is done

rh0delta:
@mixmix mistakenly made most of the pr updates in this PR https://github.com/entropyxyz/cli/pull/329, so we can merge that into this one once reviewed

src/account/utils.ts
```diff
@@ -28,11 +28,11 @@ export async function selectAndPersistNewAccount (configPath: string, newAccount
   })
 }
 
-export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress: string) {
+export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress?: string) {
   const storedConfig = await config.get(configPath)
   const { accounts } = storedConfig
 
-  const account = findAccountByAddressOrName(accounts, accountNameOrAddress)
+  const account = findAccountByAddressOrName(accounts, accountNameOrAddress || storedConfig.selectedAccount)
```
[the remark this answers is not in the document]
mixmix: @rh0delta can you either make the change or say what your thinking is? (not resolved IMO)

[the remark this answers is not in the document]
rh0delta: oh i made this change in the other PR accidentally, which is why i mentioned in our chat to review that one first since it is already being merged into this PR once ready

package.json
frankiebee: I think it's fine to merge this into dev. I would make this a blocker though for dev -> main

frankiebee:
> I like this, it's really interesting seeing how little we might need the full SDK... also begs the question "what is it even used for", "what is our policy for when we use SDK and when we use raw substrate?"
> 
you probably asking this because you have'nt needed to deal with signing. The sdk is meant to make signing and the format checking of registering with programs simple.

frankiebee (commented):
Looks good to me thanks for the work!

frankiebee (approved):

PR #327, opened 2024-12-12
mixmix - fixes to naynay/any-balance

mixmix:
found some bugs
- with calls to `getTokenDetails`
- faucet funding (`options` vs `opts`)

### DIFF
--- a/src/account/utils.ts
+++ b/src/account/utils.ts
@@ -28,11 +28,11 @@ export async function selectAndPersistNewAccount (configPath: string, newAccount
   })
 }
 
-export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress: string) {
+export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress?: string) {
   const storedConfig = await config.get(configPath)
   const { accounts } = storedConfig
 
-  const account = findAccountByAddressOrName(accounts, accountNameOrAddress)
+  const account = findAccountByAddressOrName(accounts, accountNameOrAddress || storedConfig.selectedAccount)
   if (!account) throw Error(`Unable to persist verifyingKey "${verifyingKey}" to unknown account "${accountNameOrAddress}"`)
 
   // persist to config, set selectedAccount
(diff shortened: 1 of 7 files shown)


mixmix:
Also found bugs with default account (no `--account` set)

src/transfer/command.ts
```diff
@@ -18,12 +18,12 @@ export function entropyTransferCommand () {
...
       const entropy = await loadEntropyCli(opts)
       const transferService = new EntropyTransfer(entropy, opts.endpoint)
-      const { symbol } = await getTokenDetails(entropy)
+      const { symbol } = await getTokenDetails(entropy.substrate)
 
       await transferService.transfer(destination, amount)
 
       cliWrite({
-        source: opts.account,
+        source: entropy.keyring.getAccount().admin.address,
```
mixmix, line 26: we can't use `opts.account` cos that could be
- non-existent
- an alias (and we want an address)

mixmix: changed this to match what we actually use : `entropy.accounts.registration.address`

src/account/utils.ts
```diff
@@ -28,11 +28,11 @@ export async function selectAndPersistNewAccount (configPath: string, newAccount
   })
 }
 
-export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress: string) {
+export async function persistVerifyingKeyToAccount (configPath: string, verifyingKey: string, accountNameOrAddress?: string) {
```
mixmix, line 31: We need to decide somewhere (here or upstream) that if no `--account` was passed in, that we are talking about the `selectedAccount` 

Is this the right place, or should this **require** an `accountNameOrAddress` to keep this function tight?

src/balance/command.ts
```diff
@@ -14,9 +14,9 @@ export function entropyBalanceCommand () {
   // view the balances of all accounts
   balanceCommand
     .description('Command to retrieive the balance of an account on the Entropy Network')
-    .argument('[account] <address|name>', [
```
mixmix, line 17: this syntax is wrong for `.argument()` - it was making messed up `--help`

src/account/utils.ts
rh0delta: we might want to require the account name or address here, and add logic to fail the method if no account name or address is supplied

mixmix: yeah can move the question up-chain to where this is consumed. I don't mind.

Take this PR over and do what you want I reckon @rh0delta

PR #329, opened 2024-12-16
[NayNay] Only substrate

### LINKED ISSUE #328: Remove unnecessary instantiations of entropy instance when only substrate is used
Now that we are able to load the substrate api without instantiating an entropy instance, we should now go through and update the rest of the cli and remove the unnecessary load of the entropy instance.

rh0delta:
## Related Issue(s)

- Resolves #328 

## Proposed Changes

- updated trasnfer to only need substrate rather than the entire entropy instance

## Testing

## Checklist

- [ ] I have added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Version header format: `[version] Name - year-month-day (entropy-core compatibil
 - Shared
   - updated return data displayed to user on account creation (create or import) [#311](https://github.com/entropyxyz/cli/pull/311)
   - Balance now displays the number of BITS to the nearest 4 decimal places [#306](https://github.com/entropyxyz/cli/pull/306)
+  - removed use of entropy instance from transfer flow [#329](https://github.com/entropyxyz/cli/pull/329)
 
 - CLI
   - updated balance command to take in any address, and be able to return the balance for the inputted address [#315](https://github.com/entropyxyz/cli/pull/315)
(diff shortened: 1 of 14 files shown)


mixmix (dismissed):
@rh0delta can you take this PR a bit further and show what it looks like to remove `entropy` completely from `interatction` and `command`. Right now we still have `loadEntropyCLI` or `loadEntropyTUI` happening upstream and the idea I had was that instantiatioin is where errors are handled etc. I want to see the value complete and that it works and we really can get away from loading `entropy`

mixmix:
I shoulda said this seems real promising as a path, nice observation!

rh0delta:
> @rh0delta can you take this PR a bit further and show what it looks like to remove `entropy` completely from `interatction` and `command`. Right now we still have `loadEntropyCLI` or `loadEntropyTUI` happening upstream and the idea I had was that instantiatioin is where errors are handled etc. I want to see the value complete and that it works and we really can get away from loading `entropy`

the idea here was to remove the instantiation of entropy where we can. however, transfer still requires some data coming from the entropy instance, i.e the signer pair coming from the keyring, so we are unable to completely remove the load of entropy upstream. The most that could be done currently, is the removal of the need for the entropy instance in the class itself.

rh0delta:
@mixmix may have found a way to completely remove entropy instance from transfer, both in TUI and CLI. please give it a re-review.

PR #319, opened 2024-12-03
NEXT RELEASE

mixmix:
PR for `dev` -> `main` must have:
- [ ] two QA approvals one from `js team` one from `dev rel`
- [x] comprehensive coverage of change log => https://github.com/entropyxyz/cli/pull/337

PR #338, opened 2025-01-15
mixmix/e2e cli

mixmix:
I went to smoke test the `dev` branch before merging into `main` and drew on the e2e shell script I had run, and immediately found a bug.

Decided it would be good to pull that work into a context where we can see those results in CI. This PR does that migration, but we will need a subsequent PR to fix bugs that this PR will make easy to check (see comments)


tests/global.test.mjs
```diff
@@ -1,12 +1,155 @@
...
+  const configPath = join(os.tmpdir(), `entropy-cli-${Date.now()}.e2e.json`)
+
+  async function execPlain (cmd) {
+    const opts = {
+      env: {
+        ...process.env,
+        ENTROPY_CONFIG: configPath,
+        // TODO: :fire: fix this @nayyir
+        // ENTROPY_ENDPOINT: 'dev' 
+        ENTROPY_ENDPOINT: 'ws://127.0.0.1:9944'
```
mixmix, line 33: :fire: this needs fixing in another PR (after this is merged)

When you put 'dev' in there you get errors that I highlighted here https://github.com/entropyxyz/cli/pull/319#issuecomment-2586048076

tests/global.test.mjs
```diff
@@ -1,12 +1,155 @@
...
+
+  // SIGN + VERIFY
+  {
+    const sigData = await exec('entropy sign -a naynay "some content!\nNICE&SIMPLE"')
+    // TODO: entropy sign --help is out of date
+    t.match(sigData.signature, /^0x[0-9a-f]{130,130}$/, '=> sigData.signature')
+
+    // TODO: :fire: fix this
+    // const result = await exec(`entropy verify -a naynay '${JSON.stringify(sigData)}'`)
+    // t.deepEqual(result, { isSignatureVerified: true }, '=> valid sig!')
```
mixmix, line 133: I couldn't get this to work. It's awesome to see verify here and I want this to work, but another PR - this one is focused on improving CI coverage for programmatic CLI

package.json
```diff
@@ -70,6 +69,7 @@
     "eslint": "^8.56.0",
     "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
+    "npm-run-all": "^4.1.5",
```
mixmix, line 72: this is a convenience module that make our scripts less of a nightmare
docs: https://www.npmjs.com/package/npm-run-all (you can run things in parallel or series! we could consider running some things in parallel ... e.g. `run-p test:global test:ts`... but that's fancy stuff for another time)

mixmix: note we use `Date.now()` as a uid ... that would occasionally not be unique if parallel :melting_face:

tests/global.test.mjs
```diff
@@ -1,12 +1,155 @@
...
+    t.match(verifyingKey, /^0x[0-9a-f]+$/, '=> verifyingKey')
+
+    const accounts = await exec('entropy account ls')
+    t.deepEqual(accounts[0].verifyingKeys, [verifyingKey], '=> persisted')
+  }
+
+  // SIGN + VERIFY
+  {
+    const sigData = await exec('entropy sign -a naynay "some content!\nNICE&SIMPLE"')
+    // TODO: entropy sign --help is out of date
```
mixmix, line 128:
```suggestion
```

issue created: https://github.com/entropyxyz/cli/issues/339

package.json
mixmix: yikes, look at the dependencies that this pulls in : https://github.com/entropyxyz/cli/pull/338#issuecomment-2591512281

It's a big ugly post, AND I feel like it's important and good to know the weight of such changes. I like seeing the capabilities that each module is using (e.g. filesystm access)

tests/global.test.mjs
mixmix: https://github.com/entropyxyz/cli/issues/342

tests/global.test.mjs
mixmix: fix here: https://github.com/entropyxyz/cli/pull/343

tests/global.test.mjs
mixmix: oh ... also "just working" now in this PR https://github.com/entropyxyz/cli/pull/343

rh0delta (approved):

